PR #28, opened 2015-09-04
Testcases updated to use minar scheduler instead of wait_ms. MDS serv…

ErkkiMehtala:
…er address corrected. Updated new security.h from cloud server.


yogpan01:
+1

PR #30, opened 2015-09-08
Add support for Firmware object

anttiylitokola:


yogpan01:
Update module.json as well.

PR #33, opened 2015-09-16
Change make based unit tests to use Yotta

anttiylitokola:


anttiylitokola:
@yogpan01 @AnttiKauppila can you check the changes?

PR #40, opened 2015-09-21
Fix uninitialized values in m2mreporthandler

anttiylitokola:
Fixes following error
 * IOTCLT-285 - Valgrind error when using observe and send PUT REST request


anttiylitokola:
@yogpan01 can you check this?

yogpan01:
+1

PR #45, opened 2015-09-23
Add multi instance to resource creation in test application

kuggenhoffen:
Multiple instance parameter is now passed from the command to create dynamic resource.


AnttiKauppila:
+1

PR #56, opened 2015-10-08
Add 64 byte limitation for string values, like device object resources, object names etc.

anttiylitokola:


anttiylitokola:
@yogpan01 please review

yogpan01:
LGTM, +1

PR #94, opened 2015-11-18
Fix coap response when sending GET or PUT when there is a multiple re…

anttiylitokola:
…source instances
@yogpan01 please review.


yogpan01:
LGTM, +1

PR #103, opened 2015-11-23
Response with bad request when setting lt,st or gt attributes to reso…

anttiylitokola:
…urce which is not numeral

 * Fix for defect IOTCLT-430 Write Attributes operation is done even when Resource type is not numeral


anttiylitokola:
@yogpan01

anttiylitokola:
rebuild

PR #105, opened 2015-11-27
Cancel observation if write attributes containc cancel parameter.

anttiylitokola:


anttiylitokola:
@yogpan01 Cancellling part was missing from previous change. I added it back even though it's not defined in omalwm2m specification. Maybe this can be removed once we support other cancelling methods like "Reset Message".

yogpan01:
LGTM, +1

PR #108, opened 2015-12-02
Fix for IOTCLT-459

teetak01:
Object instance ID MAX_ID 65535 is allowed, even though should
be forbidden by OMA

@anttiylitokola @yogpan01


yogpan01:
LGTM, +1

PR #109, opened 2015-12-03
Allow write_attributes to pass before observe

teetak01:
Adds a bit more flexibility in setting write_attributes/observe and
reverts earlier change. 

Could you check this: @yogpan01 @anttiylitokola

### DIFF
--- a/source/m2mbase.cpp
+++ b/source/m2mbase.cpp
@@ -179,12 +179,12 @@ void M2MBase::set_under_observation(bool observed,
 
     tr_debug("M2MBase::set_under_observation - observed: %d", observed);
     _observation_handler = handler;
-    if(observed) {
+    if(handler) {
         if (_base_type != M2MBase::ResourceInstance) {
             if(!_report_handler){
                 _report_handler = new M2MReportHandler(*this);
             }
-            _report_handler->set_under_observation(true);
+            _report_handler->set_under_observation(observed);
         }
     } else {
         if(_report_handler) {


yogpan01:
We checked the specification and our implementation together with @teetak01 and found that we can handle the cases if we revert some of the changes that @anttiylitokola did in last commit as part of optimization. LGTM, +1

anttiylitokola:
This indeed fix attributes writing but it will introduce another bug where client does not respect observable parameter at all. Can be tested with following sequence:
lwm2m-client setup --endpoint testEndpoint173446
lwm2m-client device
lwm2m-client register_object --address coap://127.0.0.1:5693
lwm2m-client object --name "13"
lwm2m-client dynamic_resource --name "37" --resource_operation 3 --observable "0"
lwm2m-client set_value --name 37 --value 2
lwm2m-client register
lwm2m-client set_value --name 37 --value 3
From the trace can be seen that client will try to send notification even it's not allowed.

We need to find another way to fix this.

yogpan01:
How about if we change this back ?
_report_handler->set_under_observation(observed); to
_report_handler->set_under_observation(true);

anttiylitokola:
That does not help. One way is to keep observable status in M2MReportHandler and check that inside set_value() function or maybe the check is better to do in upper level?
void M2MResourceInstance::report()
.....
if (report_handler && is_observable()) {
                if(_value) {
                    report_handler->set_value(atof((const char*)_value));
                } else {
                    report_handler->set_value(0);
                }
            }

yogpan01:
Also, check has to be here
M2MReportHandler *report_handler = M2MBase::report_handler();
                if(report_handler && is_observable()) {
                    report_handler->set_notification_trigger();
                }

teetak01:
I'll review this.

PR #112, opened 2015-12-07
Allow write_attributes to pass before observe

anttiylitokola:


anttiylitokola:
@teetak01, please review

PR #114, opened 2015-12-10
Fix object instance id when created using Create method.

anttiylitokola:
* Fix error IOTCLT-428 Create object instance problem


anttiylitokola:
@teetak01, @AnttiKauppila please review.

AnttiKauppila:
+1

PR #119, opened 2015-12-17
Fix IOTCLT-462 - Delete operation on Object Instance always deletes I…

anttiylitokola:
…nstance with id=0


anttiylitokola:
@teetak01

PR #120, opened 2015-12-17
Removing mDS and mbed DS references

iriark01:
We should call it mbed Device Server from now on


yogpan01:
Closing this PR as this is related with documentation update and it is already handled separately through doc review in another PR.

PR #123, opened 2015-12-17
Removing mDS and mbed DS references

iriark01:
We should call it mbed Device Server from now on


anttiylitokola:
PR is no longer valid.

PR #129, opened 2016-01-05
int64->int32

teetak01:

I would like to change most of the int64 variables to int32:s. This is for few reasons:

1) Resolves http://jira.arm.com/browse/IOTCLT-496
 - other way to resolve this ticket would require either changing the current snprintf implementation to stringstream or by changing newlib-nano to more memory consuming newlib.  (arm-mbed uses newlib-nano which does not support %lld)

I would rather use implementation that uses minimal amount of memory and space, I suspect moving to std::stringstream would result in still larger included libraries

2) We do not really need int64 size variables for device objects or firmware. Considering embedded devices, using smaller variables might give speed/footprint advantages

### DIFF
--- a/mbed-client/m2mdevice.h
+++ b/mbed-client/m2mdevice.h
@@ -112,7 +112,7 @@ public:
      * @param value, Value to be set on the resource, in Integer format.
      * @return M2MResource if created successfully, else NULL.
      */
-    M2MResource* create_resource(DeviceResource resource, int64_t value);
+    M2MResource* create_resource(DeviceResource resource, int32_t value);
 
     /**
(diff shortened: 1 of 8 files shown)


teetak01:
https://github.com/ARMmbed/mbed-client-testapp/pull/14

kuggenhoffen:
I don't think this will be ok because OMA spec defines int ranges from 8 to 64-bits and 32 and 64-bits for float. See Appendix C.

I'm not sure if this is used elsewhere but I would rather implement a custom int to string, and vice versa, function maybe?

teetak01:
Making a better implementation.

PR #131, opened 2016-01-11
Increment observation number when sending notification.

anttiylitokola:
Fixes following errros:
 * IOTCLT-492 - Observe option is not incremented in notification response
 * IOTCLT-506 - Object/Object instance Notify is plain text in first ACK


anttiylitokola:
@yogpan01

PR #132, opened 2016-01-12
implement itoa for int64 handling + unit-tests

teetak01:
Could you check this? @yogpan01 @anttiylitokola @kuggenhoffen 

I changed most snprintf functions, only leaving some which handled only small ints.

I defined BUFFER_SIZE seperately for most files, also increasing it for large buffers 20->21. The 20 was a bit too small for maximum int64 values.

New unit-tests also included in this commit.

### DIFF
--- a/mbed-client/m2mstring.h
+++ b/mbed-client/m2mstring.h
@@ -13,8 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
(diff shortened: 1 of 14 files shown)


teetak01:
The test O5030400F002 passes when run locally.

yogpan01:
rebuild

PR #133, opened 2016-01-13
When observing object, notification must contain only changed object …

anttiylitokola:
…instance

 * Fix error IOTCLT-526 - Object level observation is not sending individual notification for each changed object instance values


anttiylitokola:
@yogpan01

yogpan01:
LGTM , +1

PR #134, opened 2016-01-14
Fix for IOTCLT-525.

yogpan01:
There is now feature that registered resources can be omitted from
registration message body that is sent to the server from client.
Also, fixed is the handling of integer values in OMA objects


yogpan01:
@anttiylitokola please review

anttiylitokola:
rebuild

anttiylitokola:
rebuild

anttiylitokola:
LGTM +2

PR #141, opened 2016-01-28
Cancel observation when server response with RESET message.

anttiylitokola:
* Fix error IOTCLT-286 Observation cancelling by sending reset message does not cancel sending of notifications
 * IOTCLT-535 O5040200F071:Resources are not notified in pmax time when resource value changed during Object observation

### DIFF
--- a/mbed-client/m2mbase.h
+++ b/mbed-client/m2mbase.h
@@ -334,9 +334,22 @@ public:
      */
     virtual bool register_uri();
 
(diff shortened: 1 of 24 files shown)


anttiylitokola:
@yogpan01

anttiylitokola:
rebuild

yogpan01:
LGTM, +1

PR #142, opened 2016-01-29
Defect fix: Only latest resource change is notified in pmin time when observing object

anttiylitokola:
* IOTCLT-533 - Only latest resource change is notified in pmin time when observing object


anttiylitokola:
rebuild

anttiylitokola:
@yogpan01

yogpan01:
LGTM, +1

PR #145, opened 2016-02-09
Send response to server before calling value_updated callback.

anttiylitokola:
Fix error: IOTCLT-566 Retransmission issue when working with payload that is large than CoAP block


yogpan01:
Change the name of boolean variable from execute_callback to something like value_updated , which would indicate the flag usage.

yogpan01:
LGTM, +1

PR #146, opened 2016-02-10
Add flag to disable bootstrap feature

anttiylitokola:


anttiylitokola:
@yogpan01, can you check this?

yogpan01:
There are more functions which needs to be flagged out from M2MInterfaceImpl also from M2MNsdlInerface classes.

yogpan01:
How about the API defined in M2MInterface class ?

PR #160, opened 2016-02-26
Fix float values in write attributes

anttiylitokola:


anttiylitokola:
@yogpan01, something to check, this should fix the HW related issues.

yogpan01:
LGTM, +1

PR #164, opened 2016-03-07
Mbed trace

teetak01:

Update mbed Client to mbed-trace

### DIFF
--- a/module.json
+++ b/module.json
@@ -7,7 +7,8 @@
   "homepage": "https://github.com/ARMmbed/mbed-client",
   "license": "Apache-2.0",
   "dependencies": {
-    "mbed-client-c": "^2.0.0"
+    "mbed-client-c": "^2.0.0",
+    "mbed-trace": "~0.2.0"
   },
   "targetDependencies": {
     "arm": {
--- a/source/m2mbase.cpp
+++ b/source/m2mbase.cpp
@@ -19,7 +19,7 @@
 #include "mbed-client/m2mtimer.h"
 #include "include/m2mreporthandler.h"
 #include "include/nsdllinker.h"
-#include "ns_trace.h"
+#include "mbed-trace/mbed_trace.h"
 #include <ctype.h>
 #include <string.h>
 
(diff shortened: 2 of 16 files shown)


module.json
```diff
@@ -7,7 +7,8 @@
   "homepage": "https://github.com/ARMmbed/mbed-client",
   "license": "Apache-2.0",
   "dependencies": {
-    "mbed-client-c": "^2.0.0"
+    "mbed-client-c": "^2.0.0",
+    "mbed-trace": "ARMmbed/mbed-trace"
```
yogpan01: Why is this added like this ? It should be added as yotta module and not as HEAD . Isn't that this module is now published ?

teetak01: The version number is currently 0.2.1 so I think I have to change that to something like:

yogpan01: Why still do it like mbed-trace": "ARMmbed/mbed-trace#0.2.1 and not like
mbed-trace": "0.2.1" ?

anttiylitokola:
rebuild

teetak01:
Rebased and cleaned commits.

yogpan01:
LGTM

yogpan01:
@anttiylitokola please review

anttiylitokola:
Traces are missing when running smoketests. Does this need some changes to testapplication side to get client traces visible?

teetak01:
yeap, all my pull requests to mbed Client depend on each other + related changes to mbed-client-testapp.

After this also mbed-client-linux-example and mbed-client-quickstart, etc. needs to be updated also.

I confirmed that with recent changes to mbed-trace there is no seg. fault anymore from mbed-client-linux-example if the mbed Client dependencies are updated without changes to mbed-client-linux-example.

teetak01:
All smoketests PASS when locally linking the devel-branches to mbed-client-testapp#mbed-trace.

teetak01:
Added also a PR for linux-example https://github.com/ARMmbed/mbed-client-linux-example/pull/36

mbed-client-examples, mbed-client-example-6lowpan or mbed-client-quickstart do not seem to directly use traces, so those examples should be unaffected.

yogpan01:
rebuild

PR #165, opened 2016-03-07
Fix error IOTCLT-445

anttiylitokola:
Resource added to object instance after registration does not trigger notify operation


anttiylitokola:
@yogpan01

anttiylitokola:
rebuild

PR #167, opened 2016-03-10
Allow POST request on Resource with empty payload.

yogpan01:
This commit handles the case where there is POST request on Resource with
empty payload , there is no check for content-type and it will allow to
execute the resource without any payload.
Also, fixed is the case where right content type is set for OPAQUE resource when sending notifications.


anttiylitokola:
rebuild

PR #169, opened 2016-03-11
Fix for IOTSVS-683.

yogpan01:
This commit fixes the issue when client should send the
max age value alongwith the notification.


yogpan01:
@anttiylitokola please review.

anttiylitokola:
LGTM

PR #170, opened 2016-03-14
Introduce new API to get information to which object and object insta…

anttiylitokola:
…nce resource belongs


anttiylitokola:
@yogpan01m can you check this? I will update the unit tests if this looks fine.
Example usage in value_updated could be something like this:
M2MResource* resource = NULL;
M2MResourceInstance* res_instance = NULL;
M2MObjectInstance* obj_instance = NULL;
M2MObject* obj = NULL;
if(base) {
    switch(base->base_type()) {
        case M2MBase::Object:
            obj = (M2MObject *)base;
            printf("Value updated, object name %s\r\n",
                   obj->name().c_str());
        break;
        case M2MBase::ObjectInstance:
            obj_instance = (M2MObjectInstance *)base;
            printf("Value updated, object name %s, instance id %d\r\n",
                   obj_instance->name().c_str(), obj_instance->instance_id());
        break;
        case M2MBase::Resource: {
            resource = (M2MResource*)base;
            printf("Value updated, object name %s, object instance id %d, resource name %s\r\n",
                   resource->object_name().c_str(), resource->object_instance_id(), resource->name().c_str());
        }
        break;
        case M2MBase::ResourceInstance: {
            res_instance = (M2MResourceInstance*)base;
            printf("Value updated, object name %s, object instance id %d, resource name %s\r\n",
                   res_instance->object_name().c_str(), res_instance->object_instance_id(), res_instance->name().c_str());
        }
        break;
    }
}

PR #171, opened 2016-03-16
Update dev_man_serv_enable.md

anttiylitokola:
@yogpan01, @peknis01, can you check this change?


peknis01:
Looks ok to me. I'll do some edits anyway when we are done with the contents.

anttiylitokola:
Added example usage of value_updated function.

PR #175, opened 2016-03-23
Updating the dependency of mbed-trace to major version.

yogpan01:
@bogdanm please review


bogdanm:
LGTM. When we do this kind of update in a library module, we generally release a new **minor** version of the module.

yogpan01:
@bogdanm So, should I publish the mbed-client as 1.6.0 ?

bogdanm:
That's the idea, yes.

PR #177, opened 2016-03-23
Removing code tags from the table

iriark01:
Tables with code tags don't render on docs.mbed at the moment; easiest to remove them so that the table renders correctly. In the context of the table the distinction between the two fonts isn't very helpful anyway


iriark01:
You know what? It works fine for mbed OS. Whatever the problem is, it's not the code tags.

PR #182, opened 2016-03-29
Fix occasional unregister bug

anttiylitokola:
Fix error:
 * IOTCLT-637 - O8030000F015: Unregister timeout with local mDS


anttiylitokola:
@yogpan01

yogpan01:
LGTM

PR #183, opened 2016-03-29
Remove direct references to Linux platform implementations

TeroJaasko:
In order to port this component to nanostack, the makefile can
not compile the linux specific parts into the libmbedclient.a.


TeroJaasko:
@anttiylitokola @ErkkiMehtala @teetak01 @yogpan01

PR #187, opened 2016-03-31
Moving release notes from Docs to docs folder.

yogpan01:
@iriark01 Check this. If all the links are correct in the document ?


yogpan01:
Already handled through other PR from Irit.

PR #190, opened 2016-04-04
Moving TRACE_GROUP define from header to source files to avoid multip…

yogpan01:
…le definition error.

### DIFF
--- a/mbed-client/m2mconfig.h
+++ b/mbed-client/m2mconfig.h
@@ -22,8 +22,4 @@
 #include "mbed-client/m2mstring.h"
 using namespace m2m;
 
-#if defined(HAVE_DEBUG) || !defined(NDEBUG)
-#define TRACE_GROUP "mbed-client"
-#endif
-
 #endif // M2MCONFIG_H
--- a/source/m2mbase.cpp
+++ b/source/m2mbase.cpp
@@ -23,6 +23,8 @@
 #include <ctype.h>
 #include <string.h>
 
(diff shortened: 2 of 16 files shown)


yogpan01:
@anttiylitokola @teetak01

teetak01:
-1 With this PR I get plenty of trace group redefinition warnings. With the current version I do not observe any trace group redefinition issues at least not with mbed-client-linux-example or our internal test application.

Which error are you observing? mbed-trace has currently default trace_group definition "APPL" if no other TRACE_GROUP has been defined.

If the issue is with APPL I would see that there might be two ways to resolve it:

1) Make sure that TRACE_GROUP definitions are done before mbed-trace.h is included.
2) Remove default TRACE_GROUP specification from mbed-trace

At least I observed that mbed-trace.h is included in above files, notices that no TRACE_GROUP is defined, it defines the default one and then compiler goes ahead and discovers redefinition.

teetak01:
Default trace group is only used in mbed-trace if the code is compiled with "-d" flag or YOTTA_CFG_MBED_TRACE" is defined.

yogpan01:
TRACE_GROUP should not be defined in mbed-trace.h at the first place. Files using tracing can define there own TRACE_GROUP .That would be the right way to do it.

teetak01:
TRACE_GROUP default removed from mbed trace, still needs yotta update

yogpan01:
rebuild

PR #194, opened 2016-04-11
Size optimization

TeroJaasko:


TeroJaasko:
I'll remove the valgrind related commits (first four) from this branch once they are merged to master, please skip them, but they were needed to test this code.
@anttiylitokola @simosillankorva @yogpan01

source/m2mbase.cpp
```diff
@@ -188,20 +179,16 @@ void M2MBase::set_under_observation(bool observed,
...
 }
 
 void M2MBase::set_observation_token(const uint8_t *token, const uint8_t length)
 {
-    if(_token) {
-         free(_token);
-         _token = NULL;
-         _token_length = 0;
-    }
+     free(_token);
```
yogpan01: This is potential crash. _token will be NULL for the first time and this API call will crash.

source/m2mbase.cpp
```diff
@@ -279,10 +266,8 @@ M2MBase::Observation M2MBase::observation_level() const
 void M2MBase::get_observation_token(uint8_t *&token, uint32_t &token_length)
 {
     token_length = 0;
-    if(token) {
-        free(token);
-        token = NULL;
-    }
+    free(token);
```
yogpan01: Potential crash. What if token passed is NULL, calling free() might cause crash.

kjbracey-arm:
free(NULL) is legal.

kjbracey-arm:
Although if you're going some sort of wrapper/user-provided free, you have to be clear in documentation if their free is expected to permit freeing NULL. Implementers might not expect it, despite it being standard C free and C++ delete behaviour.

PR #196, opened 2016-04-14
Connection related changes

anttiylitokola:


source/m2mobject.cpp
```diff
@@ -45,25 +46,14 @@ M2MObject::~M2MObject()
...
-                obj_name += String(obj_inst_id);
-
-                free(obj_inst_id);
-                remove_resource_from_coap(obj_name);
-            }
-
             delete obj;
             obj = NULL;
         }
+        remove_resource_from_coap(M2MBase::name());
```
yogpan01: This will not clean the Object Instances under this Object, if Object itself is not part of the registration message. We still need to keep the ObjectInstance removal logic in place.

PR #197, opened 2016-04-14
Result of is_value_changed printed out

teetak01:
@yogpan01 @anttiylitokola 

Printing out if value was changed or not.


yogpan01:
LGTM

PR #198, opened 2016-04-14
Optimization RAM by replacing String usage with const char *

yogpan01:


kjbracey-arm:
Doesn't help that much. You've still got lots of pointers in RAM and lots of run-time calls to strlen. Would be a bit better if they were "const char * const", which gets the pointers into ROM. But simplest just to make them string literal defines, surely?

That might also make the strlen be compile-time - at the moment it has to be run-time, as the pointer might have changed what it's pointing at. I tend to use "sizeof(str) - 1" to guarantee compile time, if I know str is a string literal.

Although I'd rather you had sort of helper wrapper that actually takes zero-terminated strings as "const char *", which would avoid the cast and could include a (run-time) strlen, reducing overall code size.

kjbracey-arm:
Okay, I guess it helps quite a bit, compared to the "mallocs" the String was doing. Reduces heap usage. (So we /do/ have a significant dlmalloc user in our system).

But you should definitely at least make all the pointers const to keep them out of static RAM, even if you don't go for string literals. I can see around 1/4K of RAM for pointers you're never going to modify there.

PR #199, opened 2016-04-15
Remove the executable permission bit from sources.

TeroJaasko:
Some of the sources had x-bit set, which messes the ls output.
Fix these. No other changes.


TeroJaasko:
@anttiylitokola @simosillankorva @yogpan01

PR #200, opened 2016-04-18
Intoroduce new network related error codes.

anttiylitokola:
Introduce new network related error codes and fix object deletion from GRS list.


anttiylitokola:
@yogpan01

yogpan01:
LGTM

PR #201, opened 2016-04-18
Remove the m2mconstants.cpp from sources.mk as it was deleted from git.

TeroJaasko:


TeroJaasko:
@anttiylitokola @yogpan01

yogpan01:
+1, I had this fix but forgot to push it in :)

PR #203, opened 2016-04-20
New chapter added

peknis01:
New chapter with further reading (links to mbed Device Connector
documents). mkdocs.yml updated accordingly.

@yogpan01 please check whether this would help at all.

### DIFF
--- a/docs/further-reading.md
+++ b/docs/further-reading.md
@@ -1,6 +1,6 @@
-# Further reading
+# mbed Device Connector service
 
-## mbed Device Connector
+## Useful reading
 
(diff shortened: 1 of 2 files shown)


yogpan01:
Rather than naming it Further Read , it should be clearly stating that this section takes you to Device Service part .I think something to be visible from here as well https://www.mbed.com/en/development/software/mbed-client/

peknis01:
Any better now?

yogpan01:
@peknis01  This is fine, we should also add some text related to Device Connector at this level as well
https://www.mbed.com/en/development/software/mbed-client/ 
Do you know who can update that page ?

PR #204, opened 2016-04-21
Fix for IOTCLT-707

yogpan01:
This fix includes
- handling changes in resource observation behaviour to be stored in mbed-client-c
- Handling observation notification for Object Instance and Object level


teetak01:
+1, Tested vs. Connector with @muzkpr

muzkpr:
+1, Tested on Local MDS

PR #211, opened 2016-04-29
Remove stale binary size files.

TeroJaasko:
The files were there for the optimization phase, but they are
already stale and provide no value if they are not updated on
each commit.


TeroJaasko:
@yogpan01

yogpan01:
+1

PR #212, opened 2016-04-29
Fixes issue in push_back when capacity is full

teetak01:
Previous implementation overwrote the last item in the list when
the list was expanded (due the _index--). This now resizes the list
one step in advance, so this issue is avoided.

### DIFF
--- a/mbed-client/m2mvector.h
+++ b/mbed-client/m2mvector.h
@@ -100,7 +100,7 @@ class Vector
     }
 
     void push_back(const ObjectTemplate& x) {
-        if(_size == _capacity) {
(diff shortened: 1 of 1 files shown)


teetak01:
@yogpan01 @TeroJaasko @anttiylitokola can you check this? I debugged this with @yogpan01 

This fix should also fix all device object related tests which are failing.

anttiylitokola:
+1, we should add unit tests for this class also.

teetak01:
Added a ticket for the unittests. http://jira.arm.com/browse/IOTCLT-726

PR #217, opened 2016-05-03
Fix static resource value corruption

anttiylitokola:


anttiylitokola:
@yogpan01

yogpan01:
+1, verify with Dynamic resource as well.

anttiylitokola:
rebuild

PR #218, opened 2016-05-03
Extending API to support TLS functionality

yogpan01:
New APIs are introduces keeping backward compatibility
- set_random_number_callback()
  Application can pass their own platform specific random generation function callback
  which will be used be mbed-client-mbedtls to fetch random number, if not called, mbed-client-mbedtls
  continues to use default implementation which doesn't guarantee strong randomization.
- set_entropy_callback()
 Application can pass the entropy_cb function which contains strong callback function, Threshold and
 Entropy strength. mbed-client-mbedtls will add this as another source to entropy list apart from its
 default entropy source, which is NOT strong entropy source.

### DIFF
--- a/mbed-client/m2mconfig.h
+++ b/mbed-client/m2mconfig.h
@@ -22,4 +22,33 @@
 #include "mbed-client/m2mstring.h"
 using namespace m2m;
 
+// This is valid for mbed-client-mbedtls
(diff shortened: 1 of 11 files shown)


anttiylitokola:
rebuild

yogpan01:
rebuild

PR #226, opened 2016-05-06
Do not start observation when storing write attributes

anttiylitokola:
* Fix error IOTCLT-730 - Cancel observation, still observing resource + crash
 * Clean compiler warnings


anttiylitokola:
@yogpan01

source/m2mresourceinstance.cpp
```diff
@@ -124,22 +124,19 @@ M2MResourceInstance::ResourceType M2MResourceInstance::resource_instance_type()
...
         success = handler->parse_notification_attribute(query,
                 M2MBase::base_type(), _resource_type);
-        if (success) {
-            if ((handler->attribute_flags() & M2MReportHandler::Cancel) == 0) {
+        if(success) {
+            if (is_under_observation()) {
                 handler->set_under_observation(true);
             } else {
-                handler->set_under_observation(false);
+                handler->set_default_values();
```
yogpan01: Wrong else condition, this should be else of if(success)  loop.

source/m2mresourceinstance.cpp
```diff
@@ -124,21 +124,19 @@ M2MResourceInstance::ResourceType M2MResourceInstance::resource_instance_type()
...
+        if(success) {
+            if (is_under_observation()) {
                 handler->set_under_observation(true);
-            } else {
-                handler->set_under_observation(false);
             }
-        }
-        else {
-            handler->set_default_values();
+            else {
```
yogpan01: Still the same issue, else condition is for is_under_observation() and not for if(success)

PR #228, opened 2016-05-09
Silence most compiler warnings from m2mnsdlinterface.cpp

TeroJaasko:
The GCC 5.x complained for invalid format specifiers for [u]int32_t
parameters. Fix these by using the proper PRI*32 macro. Fix also
the warnings for unused variables.


TeroJaasko:
@anttiylitokola

anttiylitokola:
+1

PR #229, opened 2016-05-12
Fix a compilation error on armcc+trace combination

TeroJaasko:
The armcc build failed for the lack of PRId32/PRIu32
macros. It seems that it requires a separate define
to be used on C++ code before including the inttypes.h,
or the PRI-macros do not get defined.

### DIFF
--- a/source/m2mnsdlinterface.cpp
+++ b/source/m2mnsdlinterface.cpp
@@ -13,6 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// Note: this macro is needed on armcc to get the the PRI*32 macros
+// from inttypes.h in a C++ code.
+#ifndef __STDC_FORMAT_MACROS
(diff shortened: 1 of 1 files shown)


TeroJaasko:
@anttiylitokola @teetak01

source/m2mnsdlinterface.cpp
```diff
@@ -13,6 +13,13 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
+// Note: this macro is needed on armcc to get the the PRI*32 macros
+// from inttypes.h in a C++ code.
+#ifndef __STDC_FORMAT_MACROS
```
teetak01: Should here be some "if ARMCC flag also" ?

teetak01:
+1 Should work fine for both compilers.

anttiylitokola:
+1

PR #232, opened 2016-05-19
Add support for IAR

c1728p9:
Create the attribute m2m_deprecated and define this for both IAR and
GCC.  Use this attribute rather than GCC specific syntax.  This allows
mbed-client to compile for IAR.


mbed-client/m2mbase.h
```diff
@@ -21,6 +21,12 @@
 #include "mbed-client/m2mconfig.h"
 #include "mbed-client/m2mreportobserver.h"
 
+#if defined (__ICCARM__)
```
yogpan01: Can this be moved under m2mconfig.h ?

PR #235, opened 2016-05-20
Sending correct error code for Registration error.

yogpan01:
When server rejects Registration from client, client should not
try re-connection logic and inform to application about Invalid Parameters
which are passed to server from client.


yogpan01:
@anttiylitokola

anttiylitokola:
LGTM

PR #236, opened 2016-05-23
Multi instance fix

anttiylitokola:


test/mbedclient/utest/nsdlaccesshelper/test_nsdlaccesshelper.cpp
```diff
@@ -105,34 +105,34 @@ Test_NsdlAccessHelper::~Test_NsdlAccessHelper()
...
     CHECK(ptr != NULL);
-
-    delete __nsdl_interface;
-    __nsdl_interface = NULL;
     free(ptr);
+    ptr = NULL;
+    ptr = __nsdl_c_memory_alloc(UINT16_MAX+1);
+    CHECK(ptr == NULL);
+
+//free(ptr);
```
jupe: why this line is commented?

anttiylitokola:
@yogpan01

test/mbedclient/utest/nsdlaccesshelper/test_nsdlaccesshelper.cpp
anttiylitokola: Comment removed. However freeing in this case is not needed since the allocation fails in above step.

PR #242, opened 2016-06-01
Memory savings

TeroJaasko:


TeroJaasko:
@yogpan01 @anttiylitokola

yogpan01:
Looks ok. Since there is an API extension, @anttiylitokola remember to increase minor version when this change is merged and published.

PR #253, opened 2016-06-20
Adding error code for FORBIDDEN from mDS.

yogpan01:
In case the server returns error as FORBIDDEN, there should be no
re-connection logic triggered on client side.


yogpan01:
@anttiylitokola

anttiylitokola:
+1

teetak01:
rebuild

kjbracey-arm:
Surely most, or even all, 4.xx codes ("Client Error") should not retry? Generally 5.xx codes ("Server Error") should be the ones that retry.

If you don't want a global 4.xx check, at a minimum you should add the 4.xx codes that are explicitly documented in RFC 7252 as "don't repeat": Unauthorized & Bad Option.

teetak01:
rebuild

anttiylitokola:
rebuild

yogpan01:
rebuild

yogpan01:
rebuild

PR #256, opened 2016-06-27
Add new API to set value updated callback to individual object

anttiylitokola:


anttiylitokola:
@yogpan01, please review. Unit tests will be updated if this new API looks ok.

yogpan01:
@devran01 Can you please check this up and provide your feedback, if this is satisfying your use case ?

devran01:
@anttiylitokola @yogpan01 Thank you for the quick solution. I had a discussion with Tridonic and OpenAIS and they are able to implement their use cases with this change.

yogpan01:
@anttiylitokola Can you update unit tests for this ?

EduardPon:
Hello,

When can we expect this change will be deployed to receive the update via yotta?

anttiylitokola:
Hi,
Changes are now pushed to yotta. Version is 1.13.0.

PR #261, opened 2016-06-28
m2mnsdlinterface.cpp: added mutex calls to protect sn_nsdl_process_coap and sn_nsdl_exec

simosillankorva:
Added __mutex_claim before and __mutex_release after sn_nsdl_process_coap and sn_nsdl_exec as they access same resources.


simosillankorva:
rebuild

PR #262, opened 2016-06-28
m2mnsdlinterface.cpp: Free _endpoint->location_ptr before reallocation.

simosillankorva:
Free _endpoint->location_ptr before reallocation.


kjbracey-arm:
You shouldn't need the NULL test - free functions should normally be written to ignore requests to free NULL.  Standard C free(), C++ delete and ns_dyn_mem_free() all work like this.

Documentation of your memory_free doesn't actually state it in comments, but seems to be built on top of normal free().

Save a small amount of flash and line count by omitting unnecessary tests.

kuggenhoffen:
rebuild

simosillankorva:
rebuild

PR #275, opened 2016-08-22
Fix for IOTCLT-961

yogpan01:
This commit fixes issue of mapping of IP addresses for CoAP library is done.
The problem was hidden till now, because IP addresses of Bootstrap and mDS were quite different and client was accidently only comparing first 4 characters of the address hence it was identifying them as two different servers.
But now the IP addresses are similar to last 2 bytes hence the first 4 byte check always pass even though the IP addresses are different
IP Address of Bootstrap is 159.122.226.36
IP Address of mDS is 159.122.226.33
On comparing the first 4 characters now, the CoAP library assumes the requests are still coming for Bootstrap sequence.
This commit fixes this issue.


yogpan01:
@teetak01

teetak01:
+1

PR #279, opened 2016-09-07
Introduce new compile time flag for infinite reconnection loop

anttiylitokola:
By default flag is set and client will start reconnection flow over again. 
For example if  MBED_CLIENT_RECONNECTION_COUNT is 3 and MBED_CLIENT_RECONNECTION_INTERVAL is 5:
15s --> 30s --> 45s --> 15s --> 30s --> 45s and so on.


anttiylitokola:
@yogpan01, ready for review

yogpan01:
+1

anttiylitokola:
rebuild

PR #281, opened 2016-09-08
Fix error IOTCLT-866:

anttiylitokola:
* O5010203E010: Bootstrap: Server responds to initial /BS with DELETE, client returns ‘deleted’, should be ‘bad response’


yogpan01:
rebuild

yogpan01:
+1

PR #286, opened 2016-09-15
String to uint8_t* class member change for base class

karsev:
@TeroJaasko

### DIFF
--- a/mbed-client/m2mbase.h
+++ b/mbed-client/m2mbase.h
@@ -40,6 +40,9 @@ class M2MReportHandler;
  *  This class is the base class based on which all LWM2M object models
  *  can be created. This serves base class for Object, ObjectInstances and Resources.
  */
+
+
+
 class M2MBase : public M2MReportObserver {
 
 public:
@@ -225,7 +228,7 @@ public:
      * \brief Returns the object name.
      * \return Name of the object.
      */
-    virtual const String &name() const;
+    const String& name() const;
 
     /**
      * \brief Returns the object name in integer.
@@ -240,13 +243,13 @@ public:
     virtual uint16_t instance_id() const;
 
     /**
-     * \brief Returns the interface description of the object.
+     * \brief Returns a copy of the interface description object.
      * \return Description of the object.
      */
     String interface_description() const;
 
     /**
-     * \brief Returns the resource type of the object.
+     * \brief Returns a copy of the resource type object.
      * \return Resource type of the object.
      */
      String resource_type() const;
@@ -362,7 +365,7 @@ public:
     virtual void set_uri_path(const String &path);
 
     /**
-     * \brief Returns the URI path of the object.
+     * \brief Returns a copy of the URI path object.
      * \return URI path of the object.
      */
      String uri_path() const;
@@ -450,6 +453,7 @@ protected:
     */
     virtual uint8_t* alloc_copy(const uint8_t* source, uint32_t size);
 
+    char* stringdup(const char* s);
     /**
      * \brief Returns the Report Handler object.
      * \return M2MReportHandler object.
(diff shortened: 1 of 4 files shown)


kjbracey-arm:
Strings are char *, not uint8_t *. Doubly so if zero-terminated.

If you try to treat them as uint8_t *, you'll end up having to put casts everywhere.

(Okay, some existing internal APIs are already using uint8_t * for strings, but they're wrong.)

mbed-client/m2mbase.h
```diff
@@ -469,17 +469,22 @@ protected : // from M2MReportObserver
...
+    uint8_t                      *_interface_description;
     int32_t                     _name_id;
     uint32_t                    _max_age;
     uint16_t                    _instance_id;
     uint16_t                    _observation_number;
     uint8_t                     *_token;
     uint8_t                     _token_length;
+    uint8_t                     _resource_type_length;
+    uint8_t                     _uri_path_length;
+    uint8_t                     _interface_description_length;
```
TeroJaasko: Previous API did not have practical limit on these lengths. If you are using zero terminated strings, why still keep separate length field? The code is not too strict on overflow checks on length, as it has been 32bit.

source/m2mbase.cpp
```diff
@@ -114,6 +133,10 @@ M2MBase::M2MBase(const String & resource_name,
   _is_under_observation(false),
   _function_pointer(NULL)
 {
+   //_name = alloc_copy((const uint8_t*)resource_name.c_str(), resource_name.size());
+  // _name_length = resource_name.size();
```
TeroJaasko: Please do not add dead code into comments.

source/m2mbase.cpp
```diff
@@ -129,6 +152,14 @@ M2MBase::~M2MBase()
...
     free(_token);
     delete _function_pointer;
+    //free(_name);
+    //_name=NULL;
+    free(_resource_type);
+    _resource_type=NULL;
+    free(_uri_path);
+    _uri_path=NULL;
+    free(_interface_description);
+    _interface_description=NULL;
```
TeroJaasko: No need to clear pointers on destructor.

source/m2mbase.cpp
```diff
@@ -247,14 +306,24 @@ uint16_t M2MBase::instance_id() const
     return _instance_id;
 }
 
-const String& M2MBase::interface_description() const
+String M2MBase::interface_description() const
 {
-    return _interface_description;
+    if(_interface_description!=NULL) {
+        return String ((const char*)_interface_description);
+        }
```
TeroJaasko: Please do not use Symbian indendation style here.

source/m2mbase.cpp
```diff
@@ -247,14 +306,24 @@ uint16_t M2MBase::instance_id() const
...
-const String& M2MBase::resource_type() const
+String M2MBase::resource_type() const
 {
-    return _resource_type;
+    if(_resource_type!=NULL) {
+        return String ((const char*)_resource_type);
+        }
+    else {
+        return String("");
+    }
```
TeroJaasko: Instead of copypasting same sequence for empty string handling, would it better to add some utility method to String class to handle it. Unfortunately the API allows the pointer to be NULL as allocation errors are mostly ignored, so this needs to be handled everywhere.

source/m2mbase.cpp
```diff
@@ -442,12 +511,21 @@ bool M2MBase::is_integer(const String &value)
 
 void M2MBase::set_uri_path(const String &path)
 {
-    _uri_path = path;
+    free(_uri_path);
+    _uri_path = NULL;
```
TeroJaasko: This NULL assignment is pointless, as the alloc_copy() return NULL one out-of-memory case anyway.

source/m2mbase.cpp
```diff
@@ -84,6 +88,15 @@ M2MBase::M2MBase(const M2MBase& other) :
     _is_under_observation(other._is_under_observation),
     _function_pointer(NULL)
 {
+    _resource_type = alloc_copy(other._resource_type, other._resource_type_length);
+    _resource_type_length = other._resource_type_length;
+    _uri_path = alloc_copy(other._uri_path, other._uri_path_length);
+    _uri_path_length = other._uri_path_length;
+    _interface_description = alloc_copy(other._interface_description, other._uri_path_length);
+    _interface_description_length = other._interface_description_length;
+    _name = other._name;
```
TeroJaasko: Is this new _name assignment really needed here, or is it part of the name-related code which is mostly in comments?

mbed-client/m2mbase.h
```diff
@@ -243,13 +243,13 @@ class M2MBase : public M2MReportObserver {
      * \brief Returns the interface description of the object.
      * \return Description of the object.
      */
-    virtual const String& interface_description() const;
+    String interface_description() const;
```
TeroJaasko: Is the virtual removal done on purpose? Is this or other method which lost the virtual overridden in child classes?

mbed-client/m2mbase.h
```diff
@@ -243,13 +243,13 @@ class M2MBase : public M2MReportObserver {
...
 
     /**
      * \brief Returns the resource type of the object.
      * \return Resource type of the object.
      */
-    virtual const String& resource_type() const;
+     String resource_type() const;
```
TeroJaasko: This breaks API and the new behaviour (alloc a copy of data) needs to be documented on comments.

PR #287, opened 2016-09-16
M2MInterfaceImpl: add null pointer checks

anttiylitokola:


anttiylitokola:
@yogpan01

source/m2minterfaceimpl.cpp
```diff
@@ -200,44 +208,48 @@ void M2MInterfaceImpl::update_registration(M2MSecurity *security_object, const u
...
-            TRANSITION_MAP_ENTRY (EVENT_IGNORED)                // state_coap_data_received
-            TRANSITION_MAP_ENTRY (EVENT_IGNORED)                // state_processing_coap_data
-            TRANSITION_MAP_ENTRY (EVENT_IGNORED)                // state_coap_data_processed
-            TRANSITION_MAP_ENTRY (STATE_UPDATE_REGISTRATION)    // state_waiting
-        END_TRANSITION_MAP(data)
-        if(_event_ignored) {
-            _event_ignored = false;
-            if (!_reconnecting)
-                _observer.error(M2MInterface::NotAllowed);
+    if (security_object) {
```
yogpan01: Not needed , in fact this breaks the API behaviour.
update_registration can handle NULL parameter . Remove this check from update_registration() API.

kjbracey-arm:
-2 on the complete re-indent causing a squillion lines of changes. Textbook example of when and why you should use an early exit for parameter validation.

(How bad coding style it is to actually have run-time null checks is more debateable - I know I'm in the minority here on that argument.).

PR #288, opened 2016-09-20
Introduce new API to publish new objects to lwm2m server as a part of …

anttiylitokola:
…registration update.


anttiylitokola:
@yogpan01

PR #293, opened 2016-09-21
m2mstring: add new version of constructor that gets string and len

TeroJaasko:
There is need for easily constructing a string out of block of
non zero terminated memory and length. Add new overload of constructor
and unit tests for it.


TeroJaasko:
@anttiylitokola @yogpan01

PR #294, opened 2016-09-21
Introducing Configuration management through Config file

yogpan01:
This commit includes
 - Support for configuring various build time parameters throuh user supplied config file
  that can be passed at compile time through Makefile system which will work for
  non mbed-OS and non yotta build system as well.


yogpan01:
@anttiylitokola Please check if the m2mconfig.h has all the configuration flags defined.

anttiylitokola:
+1

PR #305, opened 2016-10-01
Fix most of gcc warnings from invalid trace format specifiers

TeroJaasko:
Use %PRIu32 instead of %d to print uint32_t as compiler expects.

Warnings being fixed:
---8<---8<----
[Warning] mbed_trace.h@115,90: format '%d' expects argument of type 'int',
but argument 4 has type 'uint32_t {aka long unsigned int}' [-Wformat=]


yogpan01:
rebuild

yogpan01:
rebuild

yogpan01:
rebuild

PR #308, opened 2016-10-05
Remove extra checks from register and update register states

anttiylitokola:
Allow to call unregister if state is 7 == STATE_REGISTER_RESOURCE_CREATED


anttiylitokola:
@yogpan01

PR #312, opened 2016-10-13
Remove unnecessary includes of stdio.h, cstdio and stdexp

TeroJaasko:
The stdio related headers are causing headache on some
more constrained environments. Remove those as they are not
really even needed as one can and should be using the
eg. stdint and stdlib.

### DIFF
--- a/mbed-client/m2mconfig.h
+++ b/mbed-client/m2mconfig.h
@@ -20,10 +20,10 @@
 * \brief File defining all system build time configuration used by mbed-client.
 */
 
-#include <stdlib.h>
-#include <cstring>
-#include <cstdio>
 #include "mbed-client/m2mstring.h"
+
+#include <stddef.h>
+
 using namespace m2m;
 
 /**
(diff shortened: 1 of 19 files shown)


TeroJaasko:
@yogpan01 @anttiylitokola

kjbracey-arm:
stdexp?

Including stdint.h from C++ in a public header file can be a slight trap due to the silly C99 STDC_LIMIT_MACROS thing, which some C++ compilers do.

Would generally recommend using ns_types.h to get it, as it deals with that trap (by always defining those macros).

kjbracey-arm:
BTW, what doesn't have stdio? Agree that these particular headers shouldn't be needing to include it, but we've got an awful lot of code that relies on it.

yogpan01:
rebuild

yogpan01:
rebuild

TeroJaasko:
rebuild

TeroJaasko:
rebuild

TeroJaasko:
rebuild

TeroJaasko:
rebuild

PR #324, opened 2016-11-04
Fix memory leak in deserialize_object_instances

tung7970:
til might be left unreclaimed if (TYPE_OBJECT_INSTANCE != til->_type)


anttiylitokola:
Hi,
Thanks for your finding. Our CI does not handle building if change comes through the fork so I made a new PR for this https://github.com/ARMmbed/mbed-client/pull/325.

PR #326, opened 2016-11-04
Fix Coverity findings

anttiylitokola:


teetak01:
rebuild

anttiylitokola:
@TeroJaasko

teetak01:
rebuild

anttiylitokola:
@TeroJaasko, could you check this if you just have time?

source/m2mfirmware.cpp
```diff
@@ -430,5 +430,7 @@ bool M2MFirmware::check_value_range(FirmwareResource resource, int64_t value) co
 
 void M2MFirmware::set_zero_value(M2MResource *resource)
 {
-    resource->set_value((const uint8_t*)"0", 1);
+    if(resource) {
+        resource->set_value((const uint8_t*)"0", 1);
+    }
```
jankii01: Do we need else branch here?

kjbracey-arm: Don't want the if, probably. Gives the immediate impression of being an incorrect fix. Why would you want to allow NULL inputs to this function?

jankii01: I wouldn't want, but then again I'd want a clear reason for the fialure... ;) Assert?

kjbracey-arm: Maybe. Better than hiding the NULL. Usually I'm happy enough with the segmentation fault. Release images should have asserts compiled out anyway, and when debugging segmentation faults are pretty obvious.

Anyway, the Coverity warning will have been that you're not checking for NULL here, but are somewhere else on the path - means a logical inconsistency about whether you think the pointer can be NULL or not.

This if indicates you think it can be. An assert would indicat you think it can't be.

kjbracey-arm: Good point on references. In C++ they're possibly the best way to sidestep the whole issue on the language level.

Although it's still possible to pass  `*(resource_t *) NULL` and generate the segmentation fault if you try really hard, and there would be then no way to assert against that.

TeroJaasko: You could try "assert(&res != NULL)" but IMHO there is no point to validate against malicious intent of the developer. :)

kjbracey-arm: I think the compiler is at liberty to optimise that test out. Address of an object can't be NULL.

anttiylitokola:
rebuild

jankii01:
Looks good to me.

PR #329, opened 2016-11-08
Define timer_expired as virtual in report handler

kuggenhoffen:


kuggenhoffen:
@TeroJaasko 
The ARMCC build was complaining about this.

PR #333, opened 2016-11-15
Memopt heap allocated connection security

kuggenhoffen:


kuggenhoffen:
@anttiylitokola 
Can you check this looks ok? I modified the binding mode enum so that TCP and TCP_QUEUE modes' bits match the nsdl binding mode bits for udp and upd queue. This way we don't need to do any unnecessary mapping like before.

anttiylitokola:
+2

PR #334, opened 2016-11-15
set_under_observation modified to check the observed parameter

simosillankorva:
Changed set_under_observation not to set the resource under observation if parameter observed is false.


simosillankorva:
rebuild

simosillankorva:
rebuild

simosillankorva:
rebuild

simosillankorva:
@anttiylitokola @kuggenhoffen This task is pending for review.

anttiylitokola:
+1

anttiylitokola:
This has been reverted since it caused some regression when observation happens in object level. Can be reproduced using testcase O5040300F025.
https://github.com/ARMmbed/mbed-client/commit/efe172952fbc8b26cee941233e65e7584a882904

PR #336, opened 2016-11-17
New API for getting callbacks when notification is sent to server

anttiylitokola:


anttiylitokola:
@marcuschangarm, Could you check the API, it should solve the issue described here http://jira.arm.com/browse/IOTCLT-1132

mbed-client/m2mresourceinstance.h
```diff
@@ -263,6 +264,18 @@ friend class M2MResource;
     */
     virtual M2MBlockMessage* block_message() const;
 
+    /**
+     * @brief Sets the function that is executed when this object receives
+     * response(Empty ACK) for notification message.
+     * @param callback The function pointer that is called.
+     */
+    void set_notification_sent_callback(notification_sent_callback callback);
```
marcuschangarm: I think the type should be a function pointer object so that you can both set a regular function pointer but also a member function with an object context.

Preferably there should be a templated version as well.

anttiylitokola:
rebuild

anttiylitokola:
rebuild

anttiylitokola:
rebuild

anttiylitokola:
rebuild

teetak01:
rebuild

teetak01:
rebuild

anttiylitokola:
rebuild

anttiylitokola:
rebuild

PR #341, opened 2016-11-28
Fix empty ack handling.

anttiylitokola:
Base type check was missing and it was causing a crash.


anttiylitokola:
@TeroJaasko

anttiylitokola:
rebuild

source/m2mnsdlinterface.cpp
```diff
@@ -557,8 +557,10 @@ uint8_t M2MNsdlInterface::received_from_server_callback(struct nsdl_s *nsdl_hand
                 tr_debug("M2MNsdlInterface::received_from_server_callback - Empty ACK, msg id: %d", coap_header->msg_id);
                 M2MBase *base = find_resource("", coap_header->token_ptr, coap_header->token_len);
                 if (base) {
-                    M2MResource* res = static_cast<M2MResource*> (base);
-                    res->notification_sent();
+                    if (M2MBase::Resource == base->base_type()) {
```
TeroJaasko: If this is only needed on resource, not resource instance, perhaps a comment would be in place?

TeroJaasko:
+1, nice catch for memory leaks too.

PR #342, opened 2016-11-28
Nsdl interface list accesses

kuggenhoffen:
Fix issue where it is possible to modify __nsdl_interface_list while someone else is accessing it.

### DIFF
--- a/source/nsdlaccesshelper.cpp
+++ b/source/nsdlaccesshelper.cpp
@@ -18,10 +18,7 @@
 
 #include <stdlib.h>
 
-M2MNsdlInterfaceList __nsdl_interface_list;
-
 // callback function for NSDL library to call into
-
 M2MConnectionHandler *__connection_handler = NULL;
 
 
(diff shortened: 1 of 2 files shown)


kuggenhoffen:
@anttiylitokola @TeroJaasko

kuggenhoffen:
This PR depends on https://github.com/ARMmbed/mbed-client-c/pull/96

source/nsdlaccesshelper.cpp
```diff
@@ -31,7 +28,7 @@ uint8_t __nsdl_c_callback(struct nsdl_s *nsdl_handle,
                           sn_nsdl_capab_e nsdl_capab)
 {
     uint8_t status = 0;
-    M2MNsdlInterface  *interface = get_interface(nsdl_handle);
+    M2MNsdlInterface *interface = (M2MNsdlInterface*)sn_nsdl_get_context(nsdl_handle);
```
TeroJaasko: I suppose a "assert(interface)" would make it sure the context is set properly on all cases.

TeroJaasko:
+1

jankii01:
Dependency on mbed-client-c

kuggenhoffen:
Need to wait next release where we can use new non-backward compatible mbed-client-c

kuggenhoffen:
Combined with #379 into #399

PR #344, opened 2016-11-30
Use nsdl struct in m2mbase class.

anttiylitokola:
Replace String getters with const char*
Introduce new API's for setting string values(const char*)

### DIFF
--- a/mbed-client/m2mbase.h
+++ b/mbed-client/m2mbase.h
@@ -22,12 +22,14 @@
 #include "mbed-client/m2mreportobserver.h"
 #include "mbed-client/functionpointer.h"
 #include "mbed-client/m2mstringbuffer.h"
+#include "nsdl-c/sn_nsdl.h"
+#include "nsdl-c/sn_coap_header.h"
+#include "nsdl-c/sn_nsdl_lib.h"
 
 //FORWARD DECLARATION
 struct sn_coap_hdr_;
 typedef sn_coap_hdr_ sn_coap_hdr_s;
 struct nsdl_s;
-
 struct sn_nsdl_addr_;
 typedef sn_nsdl_addr_ sn_nsdl_addr_s;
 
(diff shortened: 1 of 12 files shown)


anttiylitokola:
@TeroJaasko, @kuggenhoffen, @jankii01

mbed-client/m2mbase.h
```diff
@@ -481,34 +518,28 @@ protected : // from M2MReportObserver
 
     static bool build_path(StringBuffer<MAX_PATH_SIZE_4> &buffer, const char *s1, uint16_t i1);
 
+    char* stringdup(const char* s);
+
 private:
```
TeroJaasko: Could this be just a composite member of object, ie. not allocated separately and accessed via pointer? That would remove one point of failure on construction and also reduce heap allocation overhead.

mbed-client/m2mbase.h
```diff
@@ -254,13 +280,13 @@ class M2MBase : public M2MReportObserver {
...
      */
-    virtual const String& interface_description() const;
+    virtual const char* interface_description() const;
 
     /**
      * \brief Returns the resource type of the object.
      * \return The resource type of the object.
      */
-    virtual const String& resource_type() const;
+    virtual const char* resource_type() const;
```
kuggenhoffen: Aren't these breaking backward compatibility?

anttiylitokola: Yes, this change will break the backward compatibility. In case we want to support old API we need to keep String members as well and that will of course increase the class size.

kuggenhoffen: I was thinking that if we do this, then I suppose we can get rid of the setters with &String parameters as well?

TeroJaasko:
+1. There are still room for improvement but we can sort them out on the next commits on memory_optimizations branch and we need this in there to distribute work.

mbed-client/m2mbase.h
TeroJaasko: Nevermind, it needs to be a pointer to allow passing it from the application side also.

anttiylitokola:
Merging this in since cloud-client-example goes online now.

PR #347, opened 2016-12-08
Force publishing of device resource

jankii01:
To enable the update testing & development, it is not showing up in portal otherwise.
Ref:; Marcus Chang / IOTCLT-1284

### DIFF
--- a/source/m2mdevice.cpp
+++ b/source/m2mdevice.cpp
@@ -57,7 +57,7 @@ M2MDevice::M2MDevice()
                                                                      false);
         if(res) {
             res->set_operation(M2MBase::POST_ALLOWED);
-            res->set_register_uri(false);
+            res->set_register_uri(true);
         }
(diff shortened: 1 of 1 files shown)


TeroJaasko:
How about the 6 other places where a resource is created on this class? This change makes code behave differently depending of which of the methods is used first. And please add a some kind of comment there that this change may or may not break the specification.

anttiylitokola:
Yes, let's publish all the resources for now, mandatory and the optional ones.

jankii01:
One test case fails! The 1st test is failing. 
`16:24:40 | I0010110F062 |   fail  | GET:No response received withi |             | 174.606081 [|]`

teetak01:
rebuild

jankii01:
https://github.com/ARMmbed/mbed-client/pull/350 - replaces this.

PR #348, opened 2016-12-08
m2mbase: adapt to removal of resource_type_len

TeroJaasko:
Since the mbed-client-c does not have a separate field for
length of resource_type_ptr, ie. the resource_type_len was removed,
we need to adapt to it.


TeroJaasko:
@anttiylitokola @simosillankorva @kuggenhoffen

anttiylitokola:
+1

PR #351, opened 2016-12-09
Adapt to access param change in nsdl struct.

anttiylitokola:
Hide string setters from M2MBase class.

### DIFF
--- a/source/m2mbase.cpp
+++ b/source/m2mbase.cpp
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 #include "mbed-client/m2mbase.h"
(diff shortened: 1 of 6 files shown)


anttiylitokola:
@TeroJaasko

source/m2mbase.cpp
```diff
@@ -567,11 +582,13 @@ char* M2MBase::stringdup(const char* s)
 void M2MBase::free_resources()
 {
     if (!_is_static) {
+#ifndef MEMORY_OPTIMIZED_API
```
TeroJaasko: Wouldn't free_on_delete flag check be correct here?

TeroJaasko:
+1

PR #354, opened 2016-12-09
M2mbase tuning

TeroJaasko:
Small fixes, framework additions for String to "const char*" conversions.


TeroJaasko:
@anttiylitokola

anttiylitokola:
+1

PR #356, opened 2016-12-12
M2mbase tuning

TeroJaasko:


TeroJaasko:
@anttiylitokola

source/m2mfirmware.cpp
```diff
@@ -55,35 +55,125 @@ M2MFirmware::~M2MFirmware()
...
+
+const static M2MBase::lwm2m_parameters firmware_update_params = {
+    0, // max_age
+    0, // instance_id
+    0, // name_id
+    FIRMWARE_UPDATE, // name
+    &firmware_update_params_dynamic,
+    M2MBase::Resource // base_type
+};
+
```
anttiylitokola: Content type should be changed from tlv to text/plain. TLV is applicable only in object/object instance level or in case resource has several instances.

anttiylitokola:
+1

PR #359, opened 2016-12-12
m2mbase: remove some asserts which are too strict

TeroJaasko:
The asserts were too strict, as the _sn_resource->dynamic_resource_params
really needs to be modifiable.


TeroJaasko:
@anttiylitokola 

This fixes at least the assertions which failed on O5030500F002.

PR #362, opened 2016-12-14
Bootstrap finish ack

kuggenhoffen:
Fixes a problem where the final ACK of bootstrap sequence would not get sent for a relatively long time in some cases resulting in unexpected behaviour.


kuggenhoffen:
@TeroJaasko @anttiylitokola @jankii01 @simosillankorva

source/m2mnsdlinterface.cpp
```diff
@@ -1726,7 +1726,7 @@ void M2MNsdlInterface::handle_bootstrap_finished(sn_coap_hdr_s *coap_header,sn_n
         }
         _endpoint->endpoint_name_ptr = alloc_string_copy((uint8_t*)_endpoint_name.c_str(), _endpoint_name.length());
         _endpoint->endpoint_name_len = _endpoint_name.length();
-        _observer.bootstrap_done(_security);
+        _observer.bootstrap_wait(_security);
```
TeroJaasko: This also might deserve some comment ("// final callback bootstrap_done() is called only after the ack is sent") or something similar.

source/include/m2minterfaceimpl.h
```diff
@@ -328,6 +330,7 @@ friend class M2MInterfaceFactory;
         STATE_BOOTSTRAP,
         STATE_BOOTSTRAP_ADDRESS_RESOLVED,
         STATE_BOOTSTRAP_RESOURCE_CREATED,
+        STATE_BOOTSTRAP_WAIT,
         STATE_BOOTSTRAPPED,
         STATE_REGISTER, //5
```
anttiylitokola: Comments //5 etc. are not matching anymore.

anttiylitokola:
LGTM

kuggenhoffen:
rebuild

PR #363, opened 2016-12-15
m2minterafces: add path parameter to value_updated callbacks

TeroJaasko:
Add the path parameter containing the full resource path to
value_updated callbacks. This change is needed as the uri_path() was
removed from the M2MBase and the cloud client forwarding APIs
require that information to be available.

Note: this will break the client callback API and the client needs
to add ",const char *path" to their valued_updated() implementation
methods.


TeroJaasko:
@anttiylitokola @simosillankorva @kuggenhoffen @teetak01

PR #364, opened 2016-12-15
m2mfirmware: optionally constify the static structures

TeroJaasko:
If the old API is disabled by MEMORY_OPTIMIZED_API define,
we can put the sn_nsdl_static_resource_parameters_s structs
to flash to save some precious RAM.


TeroJaasko:
@anttiylitokola @kuggenhoffen @simosillankorva

PR #365, opened 2016-12-15
Fix for using wrong storage type for atol return value

sarahmarshy:
atol returns a long, but is currently stored in a long long. This fix uses a long to store the atol return value.

http://www.cplusplus.com/reference/cstdlib/atol/

### DIFF
--- a/source/m2mnsdlinterface.cpp
+++ b/source/m2mnsdlinterface.cpp
@@ -1097,7 +1097,7 @@ String M2MNsdlInterface::coap_to_string(uint8_t *coap_data,int coap_data_length)
 
 uint64_t M2MNsdlInterface::registration_time()
 {
-    uint64_t value = 0;
+    uint32_t value = 0;
     if(_endpoint && _endpoint->lifetime_ptr) {
         value = atol((const char*)_endpoint->lifetime_ptr);
     }


sarahmarshy:
I realize this might not be the correct solution, but the debug print was showing that the value of my registration time showed `-1778384896` vs `75` with this change.

It could be that the printf does something wonky with a long formatter and a long long value.

teetak01:
Looks sensible fix.

kjbracey-arm:
Compiler should have already been warning about this. Might be worth reviewing the build warnings or settings.

It's still not quite right printing with `"%ld"`, as uint32_t isn't a signed long, and may not even be an unsigned long. You should use `trace("%" PRIu32, my_uint32)`. Using the exact-width types sadly mandates those silly print macros. Or you could do `trace("%lu", (unsigned long) my_uint32)`, which copes if unsigned long is actually bigger than uint32_t. 

And the function is defined as returning an uint64_t. Is it really supposed to be able to read 64-bit values?

If so I realise we do have various issues with dodgy C libraries not properly supporting 64-bit printing or conversion. Generally I ignore those as just "broken" in the network layers, as it only affects trace, but in this case presumably you are looking for a working 64-bit conversion for functionality. Do I recall that mbed-client somewhere has a utility to do it manually bypassing the C library?

sarahmarshy:
> And the function is defined as returning an uint64_t. Is it really supposed to be able to read 64-bit values?

https://github.com/ARMmbed/mbed-client/pull/365/files#diff-fa5ec6477ed5daa867c809b43652396cL1102
The `atol` function returns a long, so I don't think a 64 bit return value is correct in this case.

>but in this case presumably you are looking for a working 64-bit conversion for functionality. 

I am looking to improve the print message. For now, it just prints garbage for me. Whether that is from the implicit conversion from 32 to 64 bit when returning from `atol` or when printing is undetermined. I don't really care to convert back to 64-bit for the reason stated above. It's up to the developers whether the return type of this function is changed to that of `atol` or the return value of `atol` is converted to the current return type of this function. That is a design decision, but it would be nice if the print was not garbage. The current proposed changes fix the print, but, again, whether another fix would be a more appropriate design decision I'll leave to you guys!

kjbracey-arm:
It prints garbage because one shouldn't print `uint64_t` with `%ld`. The 32-bit or 64-bit storage from `atol` isn't an issue on its own, but the fact that `atol` is signed but it goes into unsigned is potentially problematic.

The proposed change doesn't correctly fix the print - `"%ld"` is not the correct print format for `uint32_t` either. Will certainly fix it on some platforms, but potentially break it on others.

I'm hoping for more input from @TeroJaasko about what the function is actually supposed to do, as it seems flawed. I'd rather it was properly fixed. So, Tero, should this be returning signed or unsigned, and 32 or 64 bit?

yogpan01:
Fix needs to be done on printing debug logs using correct format. So, closing this PR and proposing to fix tracing issue in a separate commit.

PR #369, opened 2016-12-20
m2mbase: save memory by instantiating report handler only when needed

TeroJaasko:
Save 76 bytes (plus heap overhead) per resource if the observation is
not requested.


TeroJaasko:
@anttiylitokola @karsev

PR #370, opened 2016-12-20
Put change

TeroJaasko:
* fixes for issues found out when MEMORY_OPTIMIZED_API define is set
* Removal of data duplication at C struct allocation phase
* Use path variables directly to match the uri


TeroJaasko:
Merging likely do be done only after next release

@anttiylitokola @karsev

PR #372, opened 2016-12-23
Pass external blocwise store info through constructors. external bloc…

anttiylitokola:
…kwise store flag lives in static struct side so can't be changed at runtime.


anttiylitokola:
@TeroJaasko

PR #374, opened 2016-12-28
Revert value_updated callback changes. Uri path is now stored in m2mb…

anttiylitokola:
…ase, no need to pass extra param in callback.


anttiylitokola:
@TeroJaasko

PR #375, opened 2016-12-30
Rename delete_nsdl_resource function to remove_nsdl_resource().

anttiylitokola:
Also return error code from sn_nsdl_pop_resource().


anttiylitokola:
@TeroJaasko

TeroJaasko:
The commit message is invalid, it was delete_nsdl_resource which was renamed, not remove_nsdl_resource. It might be worth of amending as it helps grepping the git history if one is trying to find a missing method. +1 for functionality.

PR #376, opened 2017-01-02
m2mresources: remove unnecessary _object_name member

TeroJaasko:
The object_name was stored as String on each resource instance
and resource. Since we now have pointers to parent, the object
name can be queried from the parent object itself instead of
storing a copy of it. This saves 12B plus heap space for name
itself per each resource(instance).


TeroJaasko:
@anttiylitokola

PR #378, opened 2017-01-04
Fix unit test to work memopt changes

anttiylitokola:


anttiylitokola:
@yogpan01, @TeroJaasko

test/mbedclient/utest/m2mnsdlinterface/test_m2mnsdlinterface.cpp
```diff
@@ -1616,40 +1646,38 @@ void Test_M2MNsdlInterface::test_find_resource()
...
     CHECK(nsdl->find_resource("name", token, 4) != NULL);
     CHECK(nsdl->find_resource("name", token, 5) == NULL);
 
-    m2mbase_stub::int_value  = 0;
-    m2mobject_stub::base_type = M2MBase::ObjectInstance;
+    m2mbase_stub::ret_counter = 0;
     CHECK(nsdl->find_resource("name/0") != NULL);
+    m2mbase_stub::ret_counter = 0;
+    CHECK(nsdl->find_resource("name/0", inst_token, 4) != NULL);
+    m2mbase_stub::ret_counter = 0;
```
TeroJaasko: Would it be worth the trouble to test also that returned value is correct? Eg. "CHECK(nsdl->find_resource("name/0", inst_token, 4) != NULL);" could be replaced with "CHECK(nsdl->find_resource("name/0", inst_token, 4) == object_instance);".

anttiylitokola: Fixed in https://github.com/ARMmbed/mbed-client/pull/378/commits/afb4aa432a91e6d98f095fa110d75c77d5c62f1e

PR #379, opened 2017-01-05
Fix nsdlinterface to use connectionhandler instance mutex

kuggenhoffen:
Changed global __connection_handler variable to be a member of
M2MNsdlInterface and added an API to set connection handler.
This allows each M2MNsdlInterface instance use the same mutex
as the connection handler that is associated with it.


kuggenhoffen:
Has dependency on #342

kuggenhoffen:
Combined with #342 into #399

PR #380, opened 2017-01-10
Fix reconnection logic in bootstrap phase.

anttiylitokola:
Fixes error IOTCLT-1318 - Retry logic does not work correctly during bootstrap.

### DIFF
--- a/source/m2minterfaceimpl.cpp
+++ b/source/m2minterfaceimpl.cpp
@@ -61,7 +61,8 @@ M2MInterfaceImpl::M2MInterfaceImpl(M2MInterfaceObserver& observer,
   _security(NULL),
   _retry_count(0),
   _reconnecting(false),
-  _retry_timer_expired(false)
+  _retry_timer_expired(false),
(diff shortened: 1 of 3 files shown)


source/m2minterfaceimpl.cpp
```diff
@@ -61,7 +61,8 @@ M2MInterfaceImpl::M2MInterfaceImpl(M2MInterfaceObserver& observer,
   _security(NULL),
   _retry_count(0),
   _reconnecting(false),
-  _retry_timer_expired(false)
+  _retry_timer_expired(false),
+  _bootstrapped(true)
```
kuggenhoffen: Isn't this a bit confusing initiating to true, does it have to be initiated to true in the beginning when it's not actually bootstrapped?

anttiylitokola: It's set as true to make it work with connector only case where bootstrap is not performed at all. Maybe the flag name can be changed if needed.

PR #381, opened 2017-01-10
Merge memory opt to master

TeroJaasko:
All the stuff from memory_optimizations -branch excluding the pathlen change plus definition of MBED_CLIENT_C_NEW_API to enable use of new API in mbed-client-c.


TeroJaasko:
@anttiylitokola @yogpan01

PR #384, opened 2017-01-11
Helper class improvements

TeroJaasko:


TeroJaasko:
Good point, the first commit had actually broken them. Fixed, squashed and rebased against current memory_optimizations -branch.

anttiylitokola:
Unit test job in Jenkins is back online, so it can be used also to check memory_optmizations changes.
https://jenkins-internal.mbed.com/view/mbed_Client/job/mbed-client-memory-optimizations_unit_tests-master/

PR #387, opened 2017-01-13
Allow sending only if not in idle state

kuggenhoffen:


JanneKiiskila:
One unit test failure.
cloud smoke test - we have that one assert.
client smoke - 3 failures.

 O5030400E063.K64F.unknown
 I0010102F103.K64F.unknown	
 I0010113F039.K64F.unknown

anttiylitokola:
Is this going to be part of 1.1 rel?

PR #388, opened 2017-01-16
Remove unnecessary location_ptr copy.

anttiylitokola:


yogpan01:
After checking with @TeroJaasko we found that it is possible to remove the usage of _endpoint from M2MNsdlInterface class completely because mbed-client-c internally keeps the copy of the data from _endpoint structure.
I suggest you can extend this implementation to remove _endpoint from member variable and refactor the implementation to handle the parameters using local variable from stack. This will save unnecessary malloc usage from heap and reduce memory fragmentation.

anttiylitokola:
Actually it's not possible to remove _endpoint variable without an API change. _endpoint is created in create_endpoint() function and passed to nsdl in registration phase, I think it's not a good solution to combine these two phases.

PR #389, opened 2017-01-18
Helper class improvements

TeroJaasko:


TeroJaasko:
This same content is now in the memory_optimizations branch, but it can now be merged directly to master too and removed from the other branch.

@anttiylitokola @yogpan01

PR #390, opened 2017-01-18
test_m2mbase: fix valgrind warning for out of bounds memory access

TeroJaasko:
Valgrind spotted a problem of reading memory past the allocated
buffer. Fix this.


anttiylitokola:
+1

PR #393, opened 2017-01-19
m2mresource: save RAM and ROM at M2MExecuteParameter

TeroJaasko:
Change the API, use and internal storage of M2MExecuteParameter.
No need to allocate temp object from heap and copy the value just
for the duration of execute callback. This also reduces code
footprint by 120B on GCC.

Note: this changes the user facing API.

### DIFF
--- a/mbed-client/m2mresource.h
+++ b/mbed-client/m2mresource.h
@@ -295,14 +295,16 @@ class M2MResource::M2MExecuteParameter {
 private:
 
     /**
-     * \brief Constructor
+     * \brief Constructor, since there is no implementation, it prevents invalid use of it
      */
     M2MExecuteParameter();
 
-    /**
-     * Destructor
-     */
-   ~M2MExecuteParameter();
+#ifdef MEMORY_OPTIMIZED_API
+    M2MExecuteParameter(const char *object_name, const char *resource_name, uint16_t object_instance_id);
+#else
+    // This is a deprecated constructor, to be removed on next release.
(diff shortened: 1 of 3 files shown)


TeroJaasko:
Just a simple drive-by fix. We need to evaluate, if this justifies a API break. If not, lets put the code behind MEMORY_OPTIMIZED_API and leave older one there with m2m_deprecated tags.

yogpan01:
@TeroJaasko Can you modify what we discussed ?

TeroJaasko:
Moved the new version behind the MEMORY_OPTIMIZED_API flag. Now only change to API is M2MExecuteParameter::get_argument_value() is "const uint8_t*" instead of "uint8_t*". But client should not try to modify the given resource value anyway.

yogpan01:
@TeroJaasko Can you please check the unit tests ? There seems to be some compilation errors in that PR check.

JanneKiiskila:
Do we need to version the API, if the new optimized is not compatible with the older optimized?

TeroJaasko:
Fixed the unit tests, changed target to memory_optimizations branch and rebased against it. 
@JanneKiiskila; yes, technically we should have a MEMORY_OPTIMIZED_API_V2 or something, but is it really worth the trouble?

TeroJaasko:
To be rebased with master after #427 is merged.

TeroJaasko:
This needs a "git rebase master" after #498 is merged.

PR #395, opened 2017-01-30
Refactor memory allocations

tpihlaja:
Removed small allocations.

### DIFF
--- a/source/m2mtlvserializer.cpp
+++ b/source/m2mtlvserializer.cpp
@@ -16,7 +16,6 @@
 #include "include/m2mtlvserializer.h"
 #include "include/nsdllinker.h"
 #include "mbed-client/m2mconstants.h"
-
 #include <stdlib.h>
 
 #define TRACE_GROUP "mClt"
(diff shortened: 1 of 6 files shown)


source/m2mtlvserializer.cpp
```diff
@@ -200,17 +193,14 @@ uint8_t* M2MTLVSerializer::serialize_id(uint16_t id, uint32_t &size)
     } else {
         *id_ptr = id & 0xFF;
     }
-    return id_ptr;
+    return;
```
TeroJaasko: No need for "return;" without value on end of a method.

source/m2mtlvserializer.cpp
```diff
@@ -151,74 +155,76 @@ void M2MTLVSerializer::serialize_resource_instance(uint16_t id, M2MResourceInsta
...
-    uint32_t id_size = id > 255 ? 2 : 1;
-    size += id_size;
     if(id > 255) {
-        *id_ptr = (id & 0xFF00) >> 8;
-        id_ptr++;
-        *id_ptr = id & 0xFF;
-        id_ptr--;
+	size=2;
+        id_ptr[0] = (id & 0xFF00) >> 8;
+        id_ptr[1] = id & 0xFF;
```
TeroJaasko: Please fix the indendation settings on your editor.

source/m2mtlvserializer.cpp
```diff
@@ -152,87 +155,76 @@ void M2MTLVSerializer::serialize_resource_instance(uint16_t id, M2MResourceInsta
...
+    *out_ptr++=tlv_type;
+
+    int i=0;
+    while(i<id_size) {
+        *out_ptr++=id_array[i++];
     }
-    memcpy(tlv+size, tlv_type, type_length);
-    memcpy(tlv+size+type_length, id_ptr, id_size);
-    memcpy(tlv+size+type_length+id_size, length_ptr, length_size);
-    memcpy(tlv+size+type_length+id_size+length_size, value, value_length);
```
TeroJaasko: I'm not fan of hand-coding the memcpy(). If you use the memcpy(), a sane compiler will select the suitable implementation based on compilation strategy chosen by developer. eg. if size optimization is chosen, the compiler will emit a call to memcpy(), but unroll that if speed optimization was chosen. In any case, the code like this is not performance critical and smallest code footprint is usually what we want.

source/m2mtlvserializer.cpp
```diff
@@ -152,87 +155,64 @@ void M2MTLVSerializer::serialize_resource_instance(uint16_t id, M2MResourceInsta
...
-        length_size = 1;
-        length_ptr = (uint8_t*)malloc(length_size+1);
-        memset(length_ptr,0,length_size+1);
-        *length_ptr = length & 0xFF;
+        size = 1;
+        length_ptr[0] = length & 0xFF;
+    } else {
+        size=0;
     }
-    size += length_size;
```
TeroJaasko: Original code added the length to given parameter value. Wasn't this feature used by caller, or is it broken?

source/m2mtlvserializer.cpp
```diff
@@ -152,87 +155,64 @@ void M2MTLVSerializer::serialize_resource_instance(uint16_t id, M2MResourceInsta
...
+    uint8_t length_array[MAX_TLV_LENGTH_SIZE];
+    serialize_length(value_length, length_size, length_array);
 
-    tlv = (uint8_t*)malloc(size + type_length + id_size + length_size + value_length+1);
-    memset(tlv,0,size + type_length + id_size + length_size + value_length+1);
+    tlv = (uint8_t*)malloc(size + type_length + id_size + length_size + value_length);
     if(data) {
         memcpy(tlv, data, size);
+        free(data);
     }
```
TeroJaasko: While here, it might be good time to add out of memory error handling too, ie. add a return value and avoid segfaulting if malloc() returned NULL. Original code was broken as well, but now it is actually easy to handle the allocation failure.

source/m2mtlvserializer.cpp
tpihlaja: The parameter was initialised to zero just before the only call to private "serialize_length".

source/m2mtlvserializer.cpp
tpihlaja: I see that higher level serialize method already returns success information. I can make change, that malloc fail will be propagated as failure from serialize method. However that method is private, and public callers of the method don't currently handle success information. So in my understanding the change will prevent segfault, but causes silent failure and possible memory corruption later. Should we now return NULL from e.g. uint8_t* M2MTLVSerializer::serialize(M2MResource *resource, uint32_t &size) in "out of memory" situation?

tpihlaja:
rebuild

yogpan01:
@TeroJaasko Is this now fine ?

TeroJaasko:
rebuild

PR #399, opened 2017-02-01
Nsdl interface list removal and mutex

kuggenhoffen:
Removes the global nsdl_interface list and use new context parameter in sn_nsdl library to determine which interface context to work in. 

Removes the global __connection_handler variable. Instead pass connection handler instance to the nsdlinterface and use the mutex methods from there.

### DIFF
--- a/source/include/nsdlaccesshelper.h
+++ b/source/include/nsdlaccesshelper.h
@@ -19,8 +19,6 @@
 #include "include/m2mnsdlinterface.h"
 
 typedef Vector<M2MNsdlInterface  *> M2MNsdlInterfaceList;
-extern M2MNsdlInterfaceList __nsdl_interface_list;
-extern M2MConnectionHandler *__connection_handler;
 
 #ifdef __cplusplus
 extern "C" {
@@ -45,9 +43,6 @@ void *__socket_malloc( void * context, size_t size);
 void __socket_free(void * context, void * ptr);
 
 M2MNsdlInterface* get_interface(struct nsdl_s* nsdl_handle);
-
-void __mutex_claim();
-void __mutex_release();
 
 #ifdef __cplusplus
 }
--- a/source/m2mnsdlinterface.cpp
+++ b/source/m2mnsdlinterface.cpp
@@ -63,7 +63,6 @@ M2MNsdlInterface::M2MNsdlInterface(M2MNsdlObserver &observer)
   _identity_accepted(false)
 {
     tr_debug("M2MNsdlInterface::M2MNsdlInterface()");
-    __nsdl_interface_list.push_back(this);
     _sn_nsdl_address.addr_len = 0;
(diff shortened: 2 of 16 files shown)


kuggenhoffen:
@yogpan01 @anttiylitokola @TeroJaasko Can someone review, so I can merge?

anttiylitokola:
Is the nsdl side change already merged in and released to yotta and mbed-os?

source/include/nsdlaccesshelper.h
```diff
@@ -1,56 +1,51 @@
...
+#include "mbed-client/m2mconnectionhandler.h"
+#include "include/m2mnsdlinterface.h"
+
+typedef Vector<M2MNsdlInterface  *> M2MNsdlInterfaceList;
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+uint8_t __nsdl_c_callback(struct nsdl_s * nsdl_handle,
```
TeroJaasko: Eww.. did you change EOL style or how come this changed every line on file just to remove four lines?

source/m2mnsdlinterface.cpp
```diff
@@ -1649,6 +1641,25 @@ void M2MNsdlInterface::handle_bootstrap_error()
     _observer.bootstrap_error();
 }
 
+void M2MNsdlInterface::set_connection_handler(M2MConnectionHandler *connection_handler)
```
TeroJaasko: Is there really a need to set this variable later, after construction? If the connection_handler was given to constructor and preferably even stored as a reference, there would not be a need to even have those "if (_connection_handler)" checks later.

source/include/nsdlaccesshelper.h
kuggenhoffen: I'll check this

source/m2mnsdlinterface.cpp
kuggenhoffen: Probably was thinking backward compatibility by extending rather than changing the constructor but now that I think about it I think it would be better to do it that way.

kuggenhoffen:
Should be in 4.0.0, I'll check if @TeroJaasko updated the yotta module

kuggenhoffen:
Yes, looks like it is in yotta and mbed-os

source/include/nsdlaccesshelper.h
kuggenhoffen: Looks like original had windows line endings (CR LF) and this makes them unix (LF only)

source/m2mnsdlinterface.cpp
yogpan01: We dont need to worry about backward compatibility of M2MNsdlInterface class, its our internal class , we can modify the parameters the way we want, 
Tero has a good point, so if @kuggenhoffen can fix it, we can merge this PR in.

PR #343, opened 2016-11-30
Use port 0 for port randomization

JanneKiiskila:
Use port 0 for randomization, the network stack in use should do a standards compliant port randomization.

PR #400, opened 2017-02-02
Change port number to 0

simosillankorva:
This should pick a the port from the private range.

### DIFF
--- a/source/m2minterfaceimpl.cpp
+++ b/source/m2minterfaceimpl.cpp
@@ -579,7 +579,7 @@ void M2MInterfaceImpl::state_bootstrap(EventData *data)
             _observer.error(error);
         }
     } else {
-        _listen_port = (rand() % (65535-49152)) + 49152;
+        _listen_port = 0;
         _connection_handler.stop_listening();
         _connection_handler.bind_connection(_listen_port);
         _connection_handler.resolve_server_address(_server_ip_address,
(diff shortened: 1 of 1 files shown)


kjbracey-arm:
I thought that had been already been fixed in some very similar looking illogical code, by relying on the network stack doing sensible allocation (triggered by setting 0 to this port variable).

Leaving it to the stack is better, as it can avoid ports in use, whereas you might hit something by accident.

At the time this was writing, lwIP and Nanostack certainly weren't doing proper randomisation, but I think you can rely on it now (presumably the PAL/NSAPI would do randomisation if they knew they were on a broken stack).

simosillankorva:
Leaving it to the stack would be better if all the stacks would randomize the port with 0. But it may not be the case as we use also other stacks, not just lwIP and Nanostack.

If the PAL would hadle this in "broken stacks" it would be ok, but current that is prabably not the case as they dont like to do anything extra.

kjbracey-arm:
Double-check with Janne, as I have a feeling that other mbed client code (maybe examples) are already leaving it to the PAL, unless that got backed out. Be consistent.

Maybe mbed-client needs a central "choose a random port" function to cover up PAL differences.

simosillankorva:
Changed to 0 as Janne told that we are using it already in some other places. Lets hope that the PAL will handle the possible problems...

simosillankorva:
rebuild

simosillankorva:
@yogpan01 @TeroJaasko @anttiylitokola Could someone take a look at this pull request? It is quite trivial change so review should not take much time.

yogpan01:
Linked with #343 , we first get in the previous PR and then merge rest of the stuff in.

PR #401, opened 2017-02-06
Allow setting server uri with buffer and length

kuggenhoffen:
This would allow using just buffer and length to set server uri resource value, no need for temporary String variable when for example reading from configuration.


kuggenhoffen:
@anttiylitokola @yogpan01

yogpan01:
rebuild

PR #403, opened 2017-02-09
Linux port

netanelgonen:
add cmake file


mbed-client-classic.lib
```diff
@@ -1 +1 @@
-https://github.com/ARMmbed/mbed-client-classic/#42f18d977122acd4f5ad9ce141b74304b357dec2
+https://github.com/ARMmbed/mbed-client-classic/#LinuxPort
```
teetak01: No branches.

CMakeLists.txt
```diff
@@ -0,0 +1,184 @@
...
+add_definitions(-DMBED_CONF_NANOSTACK_EVENTLOOP_USE_PLATFORM_TICK_TIMER)
+
+
+FILE(GLOB MBED_CLIENT_C_SRC  
+        "${MBED_CLIENT_C_PATH}/source/libNsdl/src/*.c"
+        "${MBED_CLIENT_C_PATH}/source/libNsdl/src/include/*.h"
+        
+        "${MBED_CLIENT_C_PATH}/source/libCoap/src/*.c"
+        "${MBED_CLIENT_C_PATH}/source/libCoap/src/include/*.h"
+        )
```
TeroJaasko: Don't these belong to mbed-client-c repo itself? Why would mbed-client repo hardcode the source locations on its submodules?

Port_libs/CMakeLists.txt
```diff
@@ -0,0 +1,129 @@
+INCLUDE(CMakeForceCompiler)
+# CROSS COMPILER SETTING
+cmake_minimum_required (VERSION 2.8)
+SET(CMAKE_SYSTEM_NAME Generic)
+
+#project(linuxPortLibs)
```
TeroJaasko: What is this 100 lines of commented out stuff doing?

CMakeLists.txt
netanelgonen: the mbed-client is depened on those libraries (not only mbed-client-c) so it was decieded to to chenge any library and include all of them from a single cmake
@yogpan01

yogpan01: We agreed that mbed-client will handle all the CMakeList dependencies for the underlying repos in this level. We are trying to avoid having the need of multiple CMakelist at different repo which may not be useful without mbed Client.

TeroJaasko: Fair enough.

teetak01:
This PR should not be merged. We should still keep this branch alive to allow proper deployment of the LinuxPort components. Due to this I also swiched the mbed-client-classic to point back to LinuxPort branch.

This PR should be used for merging: https://github.com/ARMmbed/mbed-client/pull/411

teetak01:
This branch is needed until mbed-client-classic LinuxPort has been merged in. Then we can update mbed-cloud-client to point to right place.

JanneKiiskila:
@netanelgonen  @teetak01  @yogpan01 On merging policy - should we try to "shorten the string" a bit with some git rebase -i magic or just do a squash merge with GitHub?

Makefile
```diff
@@ -1,59 +0,0 @@
-#
-# Makefile for mbed Client C++ Library
```
TeroJaasko: Why is this removed?

teetak01:
@netanelgonen merged the other PR in.

TeroJaasko:
The "other PR" was this: https://github.com/ARMmbed/mbed-client/pull/411.

PR #405, opened 2017-02-10
Add support for dynamic uri query param handling.

anttiylitokola:
Uri query params are passed into CoAP option list.

### DIFF
--- a/source/m2mnsdlinterface.cpp
+++ b/source/m2mnsdlinterface.cpp
@@ -290,11 +290,11 @@ bool M2MNsdlInterface::send_register_message(uint8_t* address,
                                        false);
     bool success = false;
     if(set_NSP_address(_nsdl_handle, address, address_length, port, address_type) == 0) {
-        bool msg_sent = false;
         if (_server_address) {
-            success = msg_sent = parse_and_send_uri_query_parameters();
+            success = parse_and_send_uri_query_parameters();
         }
-        if (!msg_sent) {
+        // If URI parsing fails or there is no parameters, try again without parameters
+        if (!success) {
             success = sn_nsdl_register_endpoint(_nsdl_handle,_endpoint, NULL, 0) != 0;
         }
     }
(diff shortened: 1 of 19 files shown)


anttiylitokola:
Depends on https://github.com/ARMmbed/mbed-client-c/pull/124

source/m2mnsdlinterface.cpp
```diff
@@ -228,15 +232,43 @@ bool M2MNsdlInterface::create_bootstrap_resource(sn_nsdl_addr_s *address, const
...
-                                               address,
-                                               _endpoint,
-                                               &bootstrap_endpoint);
-        tr_debug("M2MNsdlInterface::create_bootstrap_resource - _bootstrap_id %d", _bootstrap_id);
+        // Take copy of the address, uri_query_parameters() will modify the source buffer
+        bool msg_sent = false;
+        if (_server_address) {
+            char *address_copy = M2MBase::alloc_string_copy(_server_address);
+            char* query = query_string(_server_address);
+            if (query != NULL) {
```
TeroJaasko: I guess this should check for address_copy's allocation success too.

source/m2mnsdlinterface.cpp
```diff
@@ -1683,3 +1727,27 @@ void M2MNsdlInterface::change_operation_mode(M2MObject *object, M2MBase::Operati
...
+}
+
+bool M2MNsdlInterface::parse_and_send_uri_query_parameters()
+{
+    char *address_copy = M2MBase::alloc_string_copy(_server_address);
+    char* query = query_string((const char*)_server_address);
+    bool msg_sent = false;
+    if (query != NULL) {
+        int param_count = query_param_count(query);
+        char* uri_query_params[param_count];
```
TeroJaasko: Shouldn't the param count be verified for zero? That can happen if there is no query parameter. And there should also be a unit test for that case.

source/uriqueryparser.cpp
```diff
@@ -0,0 +1,58 @@
...
+// from ARM Limited or its affiliates.
+//----------------------------------------------------------------------------
+
+#include <string.h>
+#include "include/uriqueryparser.h"
+
+char* query_string(const char* uri)
+{
+    char* query = strchr((char*)uri, '?');
+    if (query != NULL) {
```
anttiylitokola: Fixed inhttps://github.com/ARMmbed/mbed-client/pull/405/commits/46c81590a5b3b03508f9fce5e2e180fd6bf7c915

source/m2mnsdlinterface.cpp
anttiylitokola: Fixed in https://github.com/ARMmbed/mbed-client/pull/405/commits/46c81590a5b3b03508f9fce5e2e180fd6bf7c915

PR #406, opened 2017-02-14
Add function set_socket

simosillankorva:
This function is needed for PAL TLS.


kuggenhoffen:
Looks reasonable to me

PR #411, opened 2017-02-20
Linux port2

teetak01:
I did some cleanup and rebased the LinuxPort. This should be ready to merge. Please do final checkup.


Port_libs/cmake_install.cmake
```diff
@@ -0,0 +1,35 @@
...
+    message(STATUS "Install component: \"${COMPONENT}\"")
+    set(CMAKE_INSTALL_COMPONENT "${COMPONENT}")
+  else()
+    set(CMAKE_INSTALL_COMPONENT)
+  endif()
+endif()
+
+if(NOT CMAKE_INSTALL_LOCAL_ONLY)
+  # Include the install script for each subdirectory.
+  include("/home/gonen/work/proj/mbed-cloud-client-example-ref/mbed-cloud-client-example/mbed-cloud-client/mbed-client/Port_libs/mbed-trace/cmake_install.cmake")
```
JanneKiiskila: This can't be right, we can't have any includes with user name in it!

netanelgonen: you are currect,
this file should not be in the git at all, this is an auto generated file from the cmake

teetak01:
Removed some autogenerated stuff.

PR #412, opened 2017-02-21
Add timer to handle registration flow timeouts

anttiylitokola:


yogpan01:
Is this needed ? I think we have enough reconnection logic on DTLS and mbed Client side to mange reconnections.

PR #417, opened 2017-02-22
m2mbase: fix compiler warning on 64bit arch

TeroJaasko:
Use standard printf format specifier for printing a pointer value.

Fix for this warning:
--8<---8<--
warning: format ‘%x’ expects argument of type ‘unsigned int’,
but argument 4 has type ‘M2MObservationHandler*’ [-Wformat=]

### DIFF
--- a/source/m2mbase.cpp
+++ b/source/m2mbase.cpp
@@ -286,7 +286,7 @@ void M2MBase::remove_observation_level(M2MBase::Observation obs_level)
 
 void M2MBase::set_observation_handler(M2MObservationHandler *handler)
 {
-    tr_debug("M2MBase::set_observation_handler - handler: 0x%x", handler);
+    tr_debug("M2MBase::set_observation_handler - handler: 0x%p", (void*)handler);
     _observation_handler = handler;
 }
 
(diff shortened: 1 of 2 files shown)


kjbracey-arm:
`"%p"` and `(void *) handler` would be simpler. (Sadly the cast is technically required).

TeroJaasko:
True. I was just scared when armcc defines the PRIxPTR as "llx" or "x" depending on word size, not simply by "p". 
`#define PRIxPTR     __PRISCNPTR "x"`
But I can change this to "p" too.

kjbracey-arm:
That's because defining "PRIxPTR" as "p" would generate a warning - you'd have the printf code trying to load a void * from memory holding a uintptr_t, and that's undefined behaviour.

uintptr_t is some integer type that can losslessly hold a pointer value, so casting to and from uintptr_t preserves a pointer. But it doesn't necessarily have the same representation as a pointer, and isn't even necessarily the same size as a pointer. It's just a sufficiently-big integer.

So PRIxPTR does have to be something like "x" or "llx".

PR #420, opened 2017-02-27
Add mbed-client-c to mbed-client

teetak01:
Adding mbed-client-c (without mbed-coap) to mbed-client to pull in
all dependencies for applications due to coap-separation change
in mbedOS 5.4.


teetak01:
FYI: Updating also the CMakeLists.txt + mbed-client-classic.

PR #421, opened 2017-03-01
Mem opt phase 2

TeroJaasko:


TeroJaasko:
Work in progress, not ready for merge. This breaks unittests also in mbed-client-classic, will commit fixes for them after some local testing.

PR #423, opened 2017-03-02
Dynamic memory options

tpihlaja:
Discussion on handling dynamic memory allocations in mbed-client - based on prototype implementation.


tpihlaja:
Automatic checks will fail because test applications should initialize dynamic memory somewhere before mbed client classes are used.

source/include/m2mdynamicmemory.h
```diff
@@ -0,0 +1,51 @@
...
+#define M2M_DYNAMIC_MEMORY_HEAP_SIZE 20000
+class M2MDynamicMemory {
+public:
+    static int memTotal;
+    static int memCount;
+
+    void * operator new (size_t size);
+    void operator delete (void * ptr);
+    void * operator new[] (size_t size);
+    void operator delete[] (void * ptr);
```
TeroJaasko: This system needs also a way to keep the original, libc allocator in place, without touching eg. "operator new" at all.

source/include/m2mdynamicmemory.h
```diff
@@ -0,0 +1,51 @@
...
+
+    static void init(void);
+    static void init(size_t size);
+    static void init(void *heapAllocation, size_t size);
+    static void print_heap_running_statistics(void);
+    static void print_heap_overall_statistics(void);
+    static void *memory_alloc(uint16_t size);
```
TeroJaasko: Doesn't this uint16_t cause now a bunch of warnings as the client uses uin32_t and/or size_t on many places?

TeroJaasko (changes requested):

mbed-client/m2mbase.h
```diff
@@ -48,7 +49,7 @@ class M2MResource;
  *  This class is the base class based on which all LWM2M object models
  *  can be created. This serves base class for Object, ObjectInstances and Resources.
  */
-class M2MBase : public M2MReportObserver {
+class M2MBase : public M2MReportObserver, public M2MDynamicMemory {
```
tpihlaja: The namespace overload of new/delete seems to be prohibited by C++ standard. Anyways we could first make namespace changes and then try the overload in different environments. I have a separate ticket for namespace update. Maybe separate namespaces for application and core part. To me inheritance seems more clear way to do "dynamic memory options". It is a bit more brute force and maybe some complications with multiple inheritance might rise, but so far it has worked fine.

source/m2mdynamicmemory.cpp
```diff
@@ -0,0 +1,249 @@
...
+#ifdef M2M_TRACE_PRINTS
+#include <stdio.h>
+#endif
+
+#ifdef M2M_DYNMEM_LIB
+static mem_stat_t memInfo;
+#endif
+static void memory_fail_callback(heap_fail_t fail) {
+#ifdef M2M_TRACE_PRINTS
+    printf("\nM2M memory failure: %u\n", fail);
```
tpihlaja: These can be considered temporary debug prints. I should study the mbed_trace system a bit before changing to use it.

source/include/m2mdynamicmemory.h
tpihlaja: yes probably, however we should first make dynmemLIB itself more flexible and build up functionality from there. Now the library seems to work "at 16-bit".

source/include/m2mdynamicmemory.h
tpihlaja: Yes, there are no flags in header file currently. I should recheck if module builds without any configuration flag. Just introducing functions shouldn't touch them...?

source/m2mnsdlinterface.cpp
```diff
@@ -74,7 +74,7 @@ M2MNsdlInterface::M2MNsdlInterface(M2MNsdlObserver &observer, M2MConnectionHandl
     // and free functions in structure and used functions for sending
     // and receiving purposes.
     _nsdl_handle = sn_nsdl_init(&(__nsdl_c_send_to_server), &(__nsdl_c_received_from_server),
-                 &(__nsdl_c_memory_alloc), &(__nsdl_c_memory_free));
+                 &(memory_alloc), &(memory_free));
```
tpihlaja: Allright. When introducing new wrappers I tried to get rid of as many old ones as possible.

source/m2mdynamicmemory.cpp
```diff
@@ -0,0 +1,249 @@
...
+#ifdef M2M_TRACE_PRINTS
+    printf("md"); /* M2M delete */
+    print_heap_overall_statistics();
+#endif
+}
+void * M2MDynamicMemory::operator new[] (size_t size) {
+    void *tmp;
+    tmp=m2m_dyn_mem_alloc(size);
+#ifdef M2M_TRACE_PRINTS
+    memTotal+=size; memCount++;
```
tpihlaja: Ok.

source/m2mnsdlinterface.cpp
tpihlaja: In fact signatures seemed to be same:
-void* __nsdl_c_memory_alloc(uint16_t size)

Both used uint16_t type size parameter.

kjbracey-arm:
I've assume this will be properly integrated with the existing ns_dynmemlib, and you're not really intending to add a second heap manager to the build?

yogpan01:
Agreed with @kjbracey-arm , this extension should be integrated with existing dynmemlib so that other components can also utilize this feature.
@tpihlaja FYI. To begin with this new component can be part of nanostack-libservice , if directly integrating into existing dynmemlib is too big task at the moment.

kjbracey-arm:
I would assume the existing nsdynmemlib API would be easily mapped to to the updated implementation by just passing a fixed heap handle.

And then there should be some way to get that base handle for the case where your system does want mbed-client or other components to share the "main" dynmemlib heap, rather than create a separate one.

kjbracey-arm:
On the smallest platforms, it can be useful to restrict to 16 bits, even if the architecture in principle supports bigger. MSP430 is an example - it has 32-bit size_t, but we don't want/need to use that, and we know our heap(s) will be <64K.  Using size_t everywhere for all object sizes can be inefficient from a RAM point of view.

We probably want a typedef for the internal size type, to leave that possibility open. Note that presumably the internal heap implementation should use that too - as I recall it's using just "int", so this current API be giving 32-bit API but 16-bit heap on the MSP430.

There's also the potential possibility for heap objects being <64K but the heap total being >64K, which I don't think has been properly thought about.  You might want different types for heap size and individual object size.

In standard C, size_t is the appropriate size for an object, and there's no type to describe the size of a heap. In this sort of embedded heap, actually size_t is the appropriate size for the heap, but the size for the objects could be smaller.

tpihlaja:
Inner working of dynmemLIB definitely needs some work to support latest size_t API change. 64K maximum heap size sounds small to me. If that is the fact then we should keep the original implementation. With size_t we aimed for generic implementation. I guess we could still (with size_t in API) limit heap and object size somehow to keep inner working simpler and more memory efficient.

kjbracey-arm:
Using size_t is generic, but excessively fat for some of our embedded targets. Having your own specific typedef allows a build-time choice independent of the system.

We certainly want to go above 64K total for the heap - I'm not too concerned about making the heap total stats APIs 32-bit, and I thought that had already been done somewhat. So I'd be okay with those numbers just being size_t.

But for the individual heap objects, so far they've always been way less than 64K, so having a typedef for the object size which is used both in APIs and in the internal heap structure would maintain existing flexibility - we could define it as smaller than size_t on teeny platforms.

And that size type could be maybe exposed in higher application APIs too - there's a whole bunch of fixed uint16_ts in APIs that I'd like to see explanded, but I wouldn't want them to be forced to size_t on every platform.

PR #434, opened 2017-03-13
Set _pmin_exceeded = true if PMIN timer expired and clarify tracing

teetak01:
I do not think we were ever properly setting the _pmin_exceeded flag as it was in the else-branch. Also regardless if we had PMIN set or not, we always printed "sending with PMIN".


yogpan01:
@anttiylitokola Please review.

PR #437, opened 2017-03-16
Reconnection logic re-work.

yogpan01:
This commit includes
 - Changing the logic that on reconnection, client will do registration update instead of full registration.
In case, the registration update fails, client will fall back to full registration.
 - Also, the application can now try to trigger the registration update , if it doesnt want to do the full registration.
  This call may however fail if the client is not registered earlier in the mDS. The reconnection logic can kick in and
  restore the state here.
 - Updated unit tests as well.


yogpan01:
@anttiylitokola Can you please check this ?

PR #441, opened 2017-03-20
Token printing to hex array

JanneKiiskila:
It is currently printing via %*.s - but seems the tokens can be non-ASCII
and this might confuse the terminals. Better to print out as hex array.

### DIFF
--- a/source/m2mnsdlinterface.cpp
+++ b/source/m2mnsdlinterface.cpp
@@ -954,7 +954,7 @@ M2MBase* M2MNsdlInterface::find_resource(const String &object_name,
                                          uint8_t token_len)
 {
     tr_debug("M2MNsdlInterface::find_resource(object level) - name (%s)", object_name.c_str());
-    tr_debug("M2MNsdlInterface::find_resource - token (%.*s)", token_len, token);
+    tr_debug("M2MNsdlInterface::find_resource - token (%s)", tr_array(token, token_len));
     M2MBase *object = NULL;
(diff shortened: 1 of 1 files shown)


TeroJaasko:
Well, there already was PR for this in https://github.com/ARMmbed/mbed-client/pull/440 but I didn't fix all of them.

JanneKiiskila:
14:27:00 12:27:00.133 | D1 <<< DutThread: [0886][DBG ][mClt]: M2MNsdlInterface::find_resource - token (<null>)

It's always printing NULL, wonder why?

JanneKiiskila:
440 PR is printing this: 12:18:33 10:18:33.113 | D1 <<< DutThread: [0348][DBG ][mClt]: M2MNsdlInterface::find_resource - token ()

So, not null but just empty? So, is the token really empty?

kjbracey-arm:
Seems that tr_array will output <null> if the pointer is NULL and the length is zero.

Which isn't necessarily wrong, but a bit unhelpful. Generally our code tends to treat ptr=NULL, len=0 as a valid empty value, so I think if len == 0, it should output "", regardless of pointer.

Watch out for that in standard C library functions though - pointers must be valid, even if the corresponding len is 0. For example memcpy of 0 bytes to or from NULL is not permitted.

(And some compilers will take advantage of that for optimisation - if they ever see you pass a pointer to a C library function, they will optimise out any following NULL checks for it, at a high enough optimisation level).

PR #444, opened 2017-03-21
Add API to M2MSecurity whether to verify cert expiration

kuggenhoffen:
Added new APIs set_verify_cert_expiration() and verify_cert_expiration()
for passing info to security impl whether to verify certificate expiration
based on time provided by device object

Added also one trace print if during bootstrap no security object is received

### DIFF
--- a/mbed-client/m2msecurity.h
+++ b/mbed-client/m2msecurity.h
@@ -208,6 +208,17 @@ public:
      */
     ServerType server_type() const;
 
+    /**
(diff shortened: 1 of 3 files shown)


kuggenhoffen:
This is related to https://github.com/ARMmbed/mbed-client-mbed-tls/pull/67

@yogpan01 @anttiylitokola If you have better idea how to get this info into the security pimpl let me know.

yogpan01:
Not needed as of now, if we dont get time from storage then we may reconsider this PR.

PR #445, opened 2017-03-22
M2MDevice: Store integers as an array

anttiylitokola:


kjbracey-arm:
Why? If this is a workaround for broken 64-bit integer support, shouldn't that be fixed?

(Note that I have to ask this because of the lack of a meaningful commit message - it doesn't provide any information that isn't in the diff.)

anttiylitokola:
Replaced with https://github.com/ARMmbed/mbed-client/pull/446

PR #446, opened 2017-03-22
Move bytearray to int conversion one layer up to TLV parser.

anttiylitokola:
Remove array to int conversion from the m2msecurity.
Add debug line to print current time received from the BS.


source/m2mnsdlinterface.cpp
```diff
@@ -1668,12 +1668,15 @@ bool M2MNsdlInterface::validate_security_object()
         uint32_t public_key_size = _security->get_resource(M2MSecurity::PublicKey)->value_length();
         uint32_t server_key_size = _security->get_resource(M2MSecurity::ServerPublicKey)->value_length();
         uint32_t pkey_size = _security->get_resource(M2MSecurity::Secretkey)->value_length();
+        M2MDevice* dev = M2MInterfaceFactory::create_device();
+        uint32_t time = dev->resource_value_int(M2MDevice::CurrentTime);
```
kuggenhoffen: Check dev!=null?

PR #447, opened 2017-03-22
unionizing instance Id and name in M2MBase

yogpan01:


source/m2mbase.cpp
```diff
@@ -354,17 +354,26 @@ M2MBase::Operation M2MBase::operation() const
...
-    return _sn_resource->name;
+    assert(_sn_resource->identifier_int_type == false);
+    return _sn_resource->identifier.name;
 }
 
 int32_t M2MBase::name_id() const
 {
-    return _sn_resource->name_id;
+    int32_t name_id = -1;
+    if(is_integer(_sn_resource->identifier.name) && strlen(_sn_resource->identifier.name) <= MAX_ALLOWED_STRING_LENGTH) {
```
TeroJaasko: This should have a "assert(_sn_resource->identifier_int_type == false)" or just handle that case here also, as now the code may use the instance_id as a string pointer with random results.

source/m2mbase.cpp
```diff
@@ -780,15 +790,17 @@ void M2MBase::free_resources()
         free(_sn_resource->dynamic_resource_params);
     }
 
-    if (_sn_resource->free_on_delete) {
-        free(_sn_resource->name);
-        free(_sn_resource);
+    if(_sn_resource->identifier_int_type == false) {
```
TeroJaasko: This can't work, the "_sn_resource->free_on_delete" is false on structs living in ROM and the freeing will panic. The original if() needs to be left there.

source/m2mbase.cpp
yogpan01: Done

source/m2mbase.cpp
yogpan01: Done

yogpan01:
Still need to fix unit tests.

mbed-client/m2mbase.h
```diff
@@ -135,9 +135,10 @@ class M2MBase : public M2MReportObserver {
     typedef struct lwm2m_parameters {
         //add multiple_instances
         uint32_t            max_age;
-        uint16_t            instance_id; // XXX: this is not properly aligned now, need to reorder these after the elimination is done
-        int32_t             name_id;
-        char*               name; //for backwards compatibility
+        mutable union {
```
TeroJaasko: Is this mutable really needed?

source/m2mbase.cpp
```diff
@@ -91,18 +91,15 @@ M2MBase::M2MBase(const String& resource_name,
...
         }
 
-        _sn_resource->name = stringdup((char*)resource_name.c_str());
+        if((!resource_name.empty())) {            
+            _sn_resource->identifier_int_type = false;
+            _sn_resource->identifier.name = stringdup((char*)resource_name.c_str());            
+        } else {
+            tr_debug("M2MBase::M2Mbase resource name is EMPTY ===========");
+            _sn_resource->identifier_int_type = true;
+        }
```
TeroJaasko: Is identifier.instance_id left uninitialized on else-branch on purpose? On the other hand, what is the default value if one forgets to set the id later? 0?

source/m2mfirmware.cpp
```diff
@@ -226,62 +226,57 @@ static sn_nsdl_dynamic_resource_parameters_s firmware_update_result_params_dynam
...
-    0, // name_id
-    (char*)FIRMWARE_STATE, // name
+    0, //(char*)FIRMWARE_STATE
     &firmware_state_params_dynamic,
     M2MBase::Resource, // base_type
     M2MBase::INTEGER,
     false,
-    false // free_on_delete
+    false, // free_on_delete
+    false  // idenifier_int_type
```
TeroJaasko: Minor typo in idenifier_int_type, which is copied to few other places too.

source/m2mfirmware.cpp
```diff
@@ -293,23 +288,27 @@ void M2MFirmware::create_mandatory_resources()
     // todo:
     // perhaps we should have a API for batch creation of objects by using a array
     // of lwm2m_parameters.
-
+    firmware_package_params.identifier.name = (char*)FIRMWARE_PACKAGE;
```
TeroJaasko: Isn't the firmware_package_params a const struct, possibly also in ROM? I guess the instance_id could be just set to a numeric value there to avoid having a "0" -string completely.

test/mbedclient/utest/m2mresource/test_m2mresource.cpp
```diff
@@ -91,7 +91,13 @@ void Test_M2MResource::test_base_type()
 
 void Test_M2MResource::test_muliptle_instances()
 {
+    m2mbase_stub::sn_resource = (M2MBase::lwm2m_parameters_s*) malloc(sizeof(M2MBase::lwm2m_parameters_s));
+    memset(m2mbase_stub::sn_resource,0,sizeof(M2MBase::lwm2m_parameters_s));
```
TeroJaasko: Is there a reason for getting these temp lwm2m_parameter structs from heap? Why not use a local variable and set the m2mbase_stub::sn_resource to point to that? Then the libc would also find out, if the code tried to mistakenly free the given struct.
Another way to reduce code would be to use calloc().

source/m2mfirmware.cpp
yogpan01: Done.

source/m2mbase.cpp
yogpan01: Fixed

source/m2mfirmware.cpp
yogpan01: Fixed

mbed-client/m2mbase.h
yogpan01: Removed.

yogpan01:
Fixed all the unit tests and the required optimization work. Ready to merge in the branch.
@TeroJaasko FYI

TeroJaasko:
Let's merge this now to memory_optimizations to get CI coverage. All recently failing tests that were be reproduced locally are now passing.

PR #449, opened 2017-03-23
Fix unit test compilation

anttiylitokola:


yogpan01:
@anttiylitokola Is this still needed or did you fix it separately ?

PR #453, opened 2017-03-28
Minor serializer cleanup

TeroJaasko:
Minor cleanups by constifying params and using reference instead of copies in few places.


TeroJaasko:
Ah, it was the last commit which broke the tlv out-of-memory-unit test.

PR #461, opened 2017-04-04
Add new API to fetch internal endpoint name

anttiylitokola:


yogpan01:
remember to add unit test as well

PR #473, opened 2017-04-13
Fix offset for deserializer

JaniSuonpera:


yogpan01:
@JaniSuonpera Do we have an automated test case for this ?

Issue #471, opened 2017-04-10
Incorrect data size is passed in deserialize_resource_instances()


EduardPon:
In the recursive call of the deserialize_resource_instances() function, an incorrect data size is passed.

See in the functions: 
```
M2MTLVDeserializer::Error M2MTLVDeserializer::deserialize_resource_instances(const 
                                                                             uint8_t *tlv,
                                                                             uint32_t tlv_size,
                                                                             uint32_t offset,
                                                                             M2MResource &resource,
                                                                             M2MObjectInstance &object_instance,
                                                                             M2MTLVDeserializer::Operation operation,
                                                                             bool update_value)
```
It should be:
```
 if(offset < tlv_size) {
        // Remove -offset! 
        error = deserialize_resource_instances(tlv, tlv_size, offset, resource, object_instance, operation, update_value);
  }
```

See also in the function:
```
M2MTLVDeserializer::Error M2MTLVDeserializer::deserialize_resource_instances(uint8_t *tlv,
                                                                             uint32_t tlv_size,
                                                                             uint32_t offset,
                                                                             M2MResource &resource,
                                                                             M2MTLVDeserializer::Operation operation,
                                                                             bool update_value)
```

ciarmcom:
ARM Internal Ref: IOTCLT-1655

yogpan01:
Thanks for reporting !
We will check this and fix it accordingly.
@JaniSuonpera Can you please check this and add test case for this as well ?

JaniSuonpera:
Fix is now merged in master branch. Can we close issue?

EduardPon:
Yes, thanks for your support.

PR #476, opened 2017-04-18
Fix offset for deserializer

JaniSuonpera:


JaniSuonpera:
Fix for https://github.com/ARMmbed/mbed-client/issues/471

PR #482, opened 2017-04-25
Refactor resource inheritance ver2

TeroJaasko:
This PR contains rebased and updated versions of the refactor work in https://github.com/ARMmbed/mbed-client/pull/459

PR #455, opened 2017-03-28
Refactor observation handler owner

TeroJaasko:

PR #483, opened 2017-04-25
m2mbase: move the _observation_handler to m2mobject

TeroJaasko:
NOTE - this change is not ready yet. I may need to implement
the storage of level in object hierarchy where the observer
was set.

Move the _observation_handler -pointer from m2mbase to
the m2mobject class. This saves four bytes of memory
on each object instance, resource and resource instance.

### DIFF
--- a/mbed-client/m2mbase.h
+++ b/mbed-client/m2mbase.h
@@ -134,9 +134,9 @@ public:
 
     typedef struct lwm2m_parameters {
         //add multiple_instances
-        uint32_t            max_age;
+        uint32_t            max_age; // todo: add flag
         union {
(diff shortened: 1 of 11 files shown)


TeroJaasko:
This PR is having the content of https://github.com/ARMmbed/mbed-client/pull/455 just rebased on top of the recently rebased and modified refactor_resource_inheritance_ver2 -branch in https://github.com/ARMmbed/mbed-client/pull/482. 

Merging of needs to be done in correct order.

TeroJaasko:
Merging into memory_optimizations to get the CI jobs view of this.

PR #487, opened 2017-05-02
Struct realingnment

TeroJaasko:


TeroJaasko:
rebuild

TeroJaasko:
This can be rebased with master after PR #498 is in, but it is needed as a base to compile code.

PR #488, opened 2017-05-04
Add support for TLV binary integers (#463)

JaniSuonpera:
This support TLV binary formats Integer, Boolean and Time.

If resource type is INTEGER then value will convert to 64-bit binary buffer. Format is two's complement big endian word. See, OMA-TS-LightweightM2M-V1_0-20170208-A, Appendix C, data Types, Integer, TLV Format.


JaniSuonpera:
This is cherry-pick from master https://github.com/ARMmbed/mbed-client/pull/463

PR #491, opened 2017-05-05
By default, changing the COAP_BLOCKWISE option to be disabled.

yogpan01:
This change can then be used to overwrite this option consistently across different build system
using same SN_COAP_MAX_BLOCKWISE_PAYLOAD_SIZE macro.


teetak01:
@yogpan01 Could you rebase this locally?

yogpan01:
@teetak01 Better ?

PR #493, opened 2017-05-09
Use correct TLV content format number

anttiylitokola:
TLV application/vnd.oma.lwm2m+tlv 11542


ghseb:
How is the state of this issue? Using the correct content type for tlv encoded data would highly increase compatiblity to the lwm2m specification ([OMA-TS-LightweightM2M-V1_0-20170208-A](http://www.openmobilealliance.org/release/LightweightM2M/V1_0-20170208-A/OMA-TS-LightweightM2M-V1_0-20170208-A.pdf) Section 6.4).

PR #495, opened 2017-05-10
Adding new API for error description

yogpan01:
This commit introduces new error_description() API which provides human readable
text string to provide more details about the error.


source/include/m2mnsdlinterface.h
```diff
@@ -385,8 +385,16 @@ class M2MNsdlInterface : public M2MTimerObserver,
...
     */
-    void handle_bootstrap_error();
+    void handle_bootstrap_error(const char *reason);
+
+    /**
+     * @brief Handle different coap errors.
+     * @param coap_header, CoAP structure.
+     * @return Error reason.
+    */
+    const char *coap_error(sn_coap_hdr_s *coap_header);
```
TeroJaasko: This might be static method to save some flash.

source/include/m2minterfaceimpl.h
```diff
@@ -435,6 +441,7 @@ friend class M2MInterfaceFactory;
     M2MConnectionHandler        _connection_handler;
     M2MNsdlInterface            _nsdl_interface;
     M2MSecurity                 *_security;
+    char                        _error_description[64];
```
TeroJaasko: I couldn't find initialization of this buffer from constructor, which is needed as one might call error_description() without actual error happening.
The 64 should also be replaced with a constant, as it is used in few other places and having different values there may have bad results.

source/m2minterfaceimpl.cpp
```diff
@@ -601,6 +628,7 @@ void M2MInterfaceImpl::state_bootstrap(EventData *data)
         if (error != M2MInterface::ErrorNone) {
             tr_error("M2MInterfaceImpl::state_bootstrap - set error as M2MInterface::InvalidParameters");
             internal_event(STATE_IDLE);
+            strncpy(_error_description,ERROR_REASON_11, sizeof(_error_description));
```
TeroJaasko: Perhaps in future the multiple calls to strncpy() could be replaced with a helper method, which could contain a #ifdef to get rid of the code completely. That would allow linker to remove the error strings too.

source/m2mnsdlinterface.cpp
```diff
@@ -1236,6 +1236,50 @@ M2MInterface::Error M2MNsdlInterface::interface_error(sn_coap_hdr_s *coap_header
     return error;
 }
 
+const char *M2MNsdlInterface::coap_error(sn_coap_hdr_s *coap_header)
+{
+    if(coap_header) {
+        switch(coap_header->msg_code) {
+            case COAP_MSG_CODE_RESPONSE_BAD_REQUEST:
+            return COAP_ERROR_REASON_1;
```
TeroJaasko: Please use "const sn_coap_hdr_s *". This indendation looks a bit iffy.

test/mbedclient/utest/m2mnsdlinterface/test_m2mnsdlinterface.cpp
```diff
@@ -1928,3 +1930,66 @@ void Test_M2MNsdlInterface::test_get_nsdl_execution_timer()
     M2MTimer &timer = nsdl->get_nsdl_execution_timer();
     (void)timer;
 }
+
+void Test_M2MNsdlInterface::test_coap_error()
+{
+    sn_coap_hdr_s *coap_header = (sn_coap_hdr_s *)malloc(sizeof(sn_coap_hdr_s));
```
TeroJaasko: No need to use malloc here, just a local variable :)

source/m2minterfaceimpl.cpp
```diff
@@ -404,25 +415,33 @@ void M2MInterfaceImpl::socket_error(uint8_t error_code, bool retry)
     }
 #endif
 
+    char error_code_des[24];
```
jupe: and seems that there is duplicate strings so maybe those should be defined somewhere else to save a bit more memory

source/m2minterfaceimpl.cpp
```diff
@@ -841,6 +873,7 @@ void M2MInterfaceImpl::state_unregister( EventData */*data*/)
     if(!_nsdl_interface.send_unregister_message()) {
         tr_error("M2MInterfaceImpl::state_unregister : M2MInterface::NotRegistered");
```
jupe: could we use same error messages in traces as well - now same thing is declared twice. Safe also some memory ;)

yogpan01:
Closing it now.

